PR #30, opened 2020-04-02
Update scheduler dependencies

andreas-bauer:
Updated all dependencies of scheduler service and its integration test.
Some of the dependencies caused security alerts.


georg-schwarz:
Thanks Andi!

PR #44, opened 2020-04-08
43 cleanup package json

georg-schwarz:


mathiaszinnen:
I like it. What do you think about also removing the watch tasks in the package-lock.json files of the integration-test tasks? Are you using them locally or can we get rid of them?

georg-schwarz:
Done that.

Also removed nodemon dependency (overlooked that previously)

mathiaszinnen (approved):

PR #54, opened 2020-04-21
WIP: API doc via SwaggerUI

### LINKED ISSUE #42: API documentation with Swagger UI
There shall be a swagger documentation for every microservice and a common swagger ui

georg-schwarz:


georg-schwarz:
Still have to remove old documentation from README files..
But feedback is already very welcome :)

adapter/src/main/java/org/jvalue/ods/adapterservice/config/SwaggerConfig.java
```diff
@@ -0,0 +1,23 @@
...
+import springfox.documentation.spring.web.plugins.Docket;
+import springfox.documentation.swagger2.annotations.EnableSwagger2;
+
+@Configuration
+@EnableSwagger2
+public class SwaggerConfig {
+
+  @Bean
+  public Docket api() {
+    return new Docket(DocumentationType.SWAGGER_2)
```
mathiaszinnen, line 17: There is no support for OpenAPI3 yet, right?

transformation/src/index.ts
```diff
@@ -1,17 +1,42 @@
...
+
 import { TransformationEndpoint } from './transformationEndpoint'
 import VM2SandboxExecutor from './vm2SandboxExecutor'
 import JSTransformationService from './jsTransformationService'
 
+import { enableSwagger } from './swagger'
+
 const port = 8080
 
+const app: Application = express()
```
mathiaszinnen, line 16: Do we need that for swagger or did you just fix it on the fly?

scheduler/src/swagger.ts
```diff
@@ -0,0 +1,31 @@
...
+    'dist/index.js', // REST API
+    'dist/interfaces/scheduling-job.js', // model
+    'dist/interfaces/datasource-config.js', // model
+  ],
+};
+
+// initialize swagger-jsdoc
+const swaggerSpec = swaggerGen(options);
+
+export function enableSwagger(app: any) {
```
mathiaszinnen, line 26: Is it possible to export the swaggerSpec and define the endpoint in index.ts? 
I think I would prefer having all the endpoints in one place, even the meta ones describing the API itself :D

transformation/src/swagger.ts
```diff
@@ -0,0 +1,35 @@
...
+    'dist/interfaces/jobResult.js',             // model classes
+    'dist/interfaces/jobError.js',              // model classes
+    'dist/interfaces/stats.js',                 // model classes
+  ],
+};
+
+// initialize swagger-jsdoc
+const swaggerSpec = swaggerGen(options);
+
+export function enableSwagger(app: any) {
```
mathiaszinnen, line 30: See comment above in transformation/swagger.ts

adapter/src/main/java/org/jvalue/ods/adapterservice/config/SwaggerConfig.java
georg-schwarz: Actually, I think there is but it requires another library / mechanism. Probably makes sense to do so?

transformation/src/index.ts
georg-schwarz: I did it because I wanted to declare the endpoint in a different file :D

scheduler/src/swagger.ts
georg-schwarz: I'd rather pull all files with API into an `api` directory or by domain. Declaring everything in the `index.ts` is something that I consider as ok, but not on the long run. Especially as services grow (one file for config endpoints, one file for stateless functionality endpoints for example)

georg-schwarz: So would renaming it to `swaggerApi.ts` be ok for you?

transformation/src/swagger.ts
georg-schwarz: See above

scheduler/src/swagger.ts
mathiaszinnen: Agree, I also think having an api directory would be the cleanest solution. So, yes, renaming would be fine!

transformation/src/swagger.ts
mathiaszinnen: See above

adapter/src/main/java/org/jvalue/ods/adapterservice/config/SwaggerConfig.java
mathiaszinnen: I think it make sense to use the newest version of the specification if it is included in a spring boot plugin. If we need some kind of workaround or a lot of effort I would rather wait until the spring boot compatible version is updated to the new spec.

georg-schwarz:
Will go over it when the event-driven version is merged

PR #56, opened 2020-04-24
Serve UI statically & Docker cleanups

### LINKED ISSUE #48: UI: Serve as static content
Currently we use a development server in the Docker image to start the UI. The goal is to serve the UI as static resources via a webserver, like nginx.

https://daten-und-bass.io/blog/serving-a-vue-cli-production-build-on-a-sub-path-with-nginx/

Please check if this solves the issue that the UI is not served correctly on Kubenetes cluster

lunedis:
This pullrequest changes the ui-service to use a multistage-Dockerfile.

First the vue.js application gets built and the resulting static files get served by an nginx webserver.

We use a custom nginx config to redirect all ui requests to /index.html, where the vuejs-router handles them.

During development I also added a .dockerignore file. A .dockerignore file stops specific files and folders from being added to the docker build context.

Example: if you run `npm install` manually inside the /ui folder, it would then copy all of the contents of node_modules/ into the docker build context, which, depending on your system, can take a few moments. 

I therefore also added .dockerignore files to all other services. They are copies of the .gitignore files. I tested the complete system, but we should make sure everything still works.

I also unified the different Dockerfiles to use the same images and a consistent comment style.

closes #48


georg-schwarz:
Looks really good to me! Good work!

One thing I noticed: navigating to e.g. `/datasources` works now, but not to `datasources/new`.
So I guess therefore, the issue is somewhere else? Any guesses?

georg-schwarz (approved):
Nice work!

PR #71, opened 2020-05-08
Show only master build state as badge

andreas-bauer:
At the moment the status badge will show very often that the build fail because it failed in one of the dev branches, even the CI in the master branch runs successfully.

With this PR it will only show the state of the master branch.


georg-schwarz (approved):
Thanks!

PR #75, opened 2020-05-11
74 covid example

### LINKED ISSUE #74: Try to import data from covid19 api (https://covid19api.com/)
Add resulting configuration to example requests in doc folder

mathiaszinnen:
closes #74


doc/example-requests/adapter.http
```diff
@@ -9,6 +9,7 @@ GET {{baseURL}}/formats  HTTP/1.1
 Content-Type: application/json
 
 ### Perform Data Import JSON
+## DEPRECATED
```
georg-schwarz: Why? Don't we still have that endpoint for preview purposes?

doc/example-requests/covid19-api/pipeline.http
```diff
@@ -0,0 +1,30 @@
...
+@dataSourceId = "1"
+
+### Store data without modification and trigger a firebase notification
+POST {{baseURL}}/pipelines  HTTP/1.1
+Content-Type: application/json
+
+{
+  "datasourceId": {{dataSourceId}},
+  "transformation":  {
+    "func": "return data;"
```
georg-schwarz: Do we need any transformations to improve data quality?

doc/example-requests/adapter.http
mathiaszinnen: Right, I forgot about that :D

doc/example-requests/covid19-api/pipeline.http
mathiaszinnen: Yes, that would be nice to show possible applications for the transformation but it is a little complicated to implement because, depending on the related datasource id, the data looks very different. I could merge the pipeline and adapter example request files for the covid example and group them together, so it is clear which pipeline relates to which datasource.

mathiaszinnen: I created three separate files now to show three example applications of the transformation functions

georg-schwarz:
Fixed typo in the filename (cointries -> countries)

georg-schwarz (approved):

PR #77, opened 2020-05-13
Merge nofitcation segregation

georg-schwarz:


georg-schwarz:
Bitte sorgfältig durchsehen. Habe noch Integrationstests, CI, Systemtests, etc... angepasst und einige Sachen noch aufgeräumt.

docker-compose.it.yml
```diff
@@ -55,16 +58,28 @@ services:
...
 
   transformation-it:
     image: open-data-service/ods-main/transformation-it
     build:
       context: ./transformation/integration-test/
     environment:
       TRANSFORMATION_API: http://transformation:8080
-      MOCK_RECEIVER_HOST: transformation-it
+
+   # ----------------- NOTIFICATION SERVICE (PORTS #TODO) --------------------
```
mathiaszinnen, line 69: Are these port informations still used somewhere, @georg-schwarz ? As far as I know they are outdated and can be removed anyway (in another issue).

notification/README.md
```diff
@@ -0,0 +1,105 @@
+# Open Data Service - Notification-Service
+
+## Getting started
+
+Make sure they keyclone container is running (see `../auth`).
```
mathiaszinnen: keycloak ;-)

mathiaszinnen: I see this typo is copied from transformation service. So lets leave it like this and fix it everywhere in a separate issue

notification/keycloak.json
```diff
@@ -0,0 +1,8 @@
+{
+    "realm": "ods-userservice",
+    "bearer-only": true,
+    "auth-server-url": "http://localhost:8080/auth",
+    "ssl-required": "external",
+    "resource": "notfication-service",
```
mathiaszinnen: typo

notification/src/interfaces/notificationService.ts
```diff
@@ -0,0 +1,6 @@
+import { NotificationRequest } from './notificationRequest'
```
mathiaszinnen, line 1: In my opinion, we should leave out the extra abstraction of the notification service interface and keep it simple. Unlike for the transformation service, I don't think there is a use case here for adding a different implementation of the notification service. What do you think @georg-schwarz?

notification/src/jsNotificationService.test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+      const expectedObject: SlackCallback = {
+        text: `Pipeline ${request.pipelineName}(${request.pipelineId}) has new data available. Fetch at ${request.dataLocation}.`
+      }
+      expect(post).toHaveBeenCalledTimes(1)
+      const expectedUrl = `https://hooks.slack.com/services/${request.workspaceId}/${request.channelId}/${request.secret}`
+      expect(post.mock.calls[0][0]).toEqual(expectedUrl)
+      expect(post.mock.calls[0][1]).toEqual(expectedObject)
+    })
+  })
```
mathiaszinnen, line 113: There is a unit test for firebase notifications missing, but I am the one to blame here, since it did not exist before ;-).
We should create another issue for that.

notification/src/vm2SandboxExecutor.ts
```diff
@@ -0,0 +1,38 @@
...
+      expression +
+      '};f(' +
+      JSON.stringify(data) +
+      ');'
+
+    let result = false
+    try {
+      result = this.vm.run(wrapper)
+    } catch (err) {
+      // console.error('Malformed expression received: ' + expression, err)
```
mathiaszinnen: uncommenting this and logging the error will probably be helpful

notification/src/vm2SandboxExecutor.ts
```diff
@@ -0,0 +1,38 @@
...
+    }
+    if (typeof result !== 'boolean') {
+      return false
```
mathiaszinnen: maybe we should log something here, too

scheduler/integration-test/src/scheduler.test.js
```diff
@@ -54,7 +59,7 @@ describe('Scheduler', () => {
   })
 
   test('GET /jobs', async () => {
-    await sleep(2000) // wait until scheduler does sync
+    await sleep(4000) // wait until scheduler does sync
```
mathiaszinnen, line 62: This is not actually related to the notification service right? Did this test fail with the shorter delay?

transformation/integration-test/package.json
```diff
@@ -3,8 +3,6 @@
   "version": "1.0.0",
   "description": "Transformation Service Integration-Tests",
   "scripts": {
-    "mock:receiver": "node src/mock.receiver.js",
-    "pretest": "npm run mock:receiver &",
     "test": "jest",
     "posttest": "killall node",
```
mathiaszinnen: the posttest script can be removed, too. No node mock servers to kill left.

transformation/integration-test/package.json
```diff
@@ -3,8 +3,6 @@
...
     "lint": "./node_modules/.bin/eslint src --ext .js --fix"
```
mathiaszinnen, line 7: koa dependencies can be removed too

transformation/integration-test/src/transformation.test.js
```diff
@@ -4,18 +4,13 @@ const waitOn = require('wait-on')
...
 const URL = process.env.TRANSFORMATION_API || 'http://localhost:8080'
 
-const MOCK_RECEIVER_PORT = process.env.MOCK_RECEIVER_PORT || 8081
-const MOCK_RECEIVER_HOST = process.env.MOCK_RECEIVER_HOST || 'localhost'
-const MOCK_RECEIVER_URL = 'http://' + MOCK_RECEIVER_HOST + ':' + MOCK_RECEIVER_PORT
-
-describe('Scheduler', () => {
-  console.log('Scheduler-Service URL= ' + URL)
+describe('Transformation Service', () => {
+  console.log('Trnasformation-Service URL= ' + URL)
```
mathiaszinnen: typo

docker-compose.it.yml
georg-schwarz: Do you mean the TRANSFORMATION_API? That's still used to separate local development env from Docker env.

notification/src/interfaces/notificationService.ts
georg-schwarz: Doesn't hurt either.... But I agree, we could refactor a little... Also a directory called `interfaces` doesn't look appealing for me, rather structure after domain (code-executor, notification, etc..)
Makes also sense to have files for each different notification type.

notification/src/vm2SandboxExecutor.ts
georg-schwarz: Throw error instead, so there is log output

scheduler/integration-test/src/scheduler.test.js
georg-schwarz: Jep

notification/src/jsNotificationService.test.ts
georg-schwarz: Opened #82

notification/src/vm2SandboxExecutor.ts
georg-schwarz: Issue was that there is no logger installed I think. See conversation below! (throwing error now)

docker-compose.it.yml
mathiaszinnen: No, I mean the port ranges that are specified for each microservice. I was thinking we don't need them anymore since docker-internal addresses are handled by traefik, but I did not know if you were using them locally somewhere ..

notification/src/vm2SandboxExecutor.ts
mathiaszinnen: perfect

mathiaszinnen (approved):

mathiaszinnen:
closes #80

PR #90, opened 2020-06-03
70 restructure readme

### LINKED ISSUE #70: Restructure README / documentation
I think our README could need a little more structuring.
Let's collect here first what should be part of the top-level README and what will go to other files, like the service READMEs..

A little inspiration: https://github.com/gentics/mesh/blob/dev/README.md

### LINKED ISSUE #73: Update README to new UI
The pictures in the howTo show an outdated UI. We need to update the pictures and description to match the current workflow for pipeline creation (Datasource and Pipeline separated).

mathiaszinnen:
Closes #70, #73 

I also updated the walkthrough for pipeline configuration via the UI. Unfortunately, the image links will not work before the pr is merged (they reference the master branch). I did not want to reference the feature branch, because otherwise, we would need to change it again after merging.


georg-schwarz (approved):
Sweet. I like it!

PR #102, opened 2020-06-12
Added Storage for services and amqp communication for notification and transformation

ke45xumo:


notification/src/models/notificationConfig.ts
```diff
@@ -0,0 +1,72 @@
+import { Entity, PrimaryColumn, Column, OneToOne, JoinColumn, PrimaryGeneratedColumn } from 'typeorm';
+import "reflect-metadata"
```
georg-schwarz: Should be unnecessary because imported in index(?)

transformation/src/interfaces/TransormationConfig.ts
```diff
@@ -0,0 +1,18 @@
+import { Entity, PrimaryGeneratedColumn, Column } from 'typeorm';
+import "reflect-metadata"
```
georg-schwarz: Should be unnecessary because imported in index(?)

notification/notification-db/Dockerfile
```diff
@@ -0,0 +1,52 @@
+FROM postgres:latest
```
georg-schwarz: File can be deleted because unused (?)

notification/package.json
```diff
@@ -11,16 +11,23 @@
         "test": "jest"
     },
     "dependencies": {
+        "@types/pg": "^7.14.3",
+        "amqplib": "^0.5.6",
         "axios": "^0.19.2",
         "body-parser": "^1.19.0",
         "cors": "^2.8.5",
+        "es6-shim": "^0.35.5",
```
georg-schwarz: Can we instead add `"lib": ["es5"  , "dom"  ,  "es6"]` to `tsconfig.json` and remove the shim?

notification/package.json
```diff
@@ -11,16 +11,23 @@
         "test": "jest"
     },
     "dependencies": {
+        "@types/pg": "^7.14.3",
```
georg-schwarz: Move to devDependencies

notification/package.json
```diff
@@ -32,14 +39,14 @@
         "concurrently": "^5.2.0",
         "eslint": "^6.8.0",
         "eslint-config-standard": "^14.1.1",
-        "eslint-plugin-import": "^2.21.2",
+        "eslint-plugin-import": "^2.20.2",
```
georg-schwarz, line 41: to 2.21.2

notification/package.json
```diff
@@ -32,14 +39,14 @@
...
         "eslint-plugin-node": "^11.1.0",
         "eslint-plugin-promise": "^4.2.1",
         "eslint-plugin-standard": "^4.0.1",
         "jest": "^25.5.4",
         "ts-jest": "^25.5.1",
         "tslib": "^1.13.0",
-        "typescript": "^3.9.5"
+        "typescript": "^3.9.3"
```
georg-schwarz: to 3.9.5

notification/src/interfaces/sandboxExecutor.ts
```diff
@@ -1,3 +1,3 @@
 export default interface SandboxExecutor {
-  evaluate(expression: string, data: object): boolean;
+  evaluate(expression: string, data: object | undefined): boolean;
```
georg-schwarz: Why also accept undefined? Is there a design decision to make?

notification/src/interfaces/transformationResults/transformationEventInterface.ts
```diff
@@ -0,0 +1,22 @@
+
+import JobResult from './jobResult';
+
+/**
+ * Event send by the Transformation service upon transformation finish
+ */
+export interface TransformationEventInterface {
+    pipelineID: number      // ID of the pipeline
```
georg-schwarz: Please use `Id` (lower camel case) instead of `ID` for consistency through the whole ODS

notification/src/jsNotificationService.ts
```diff
@@ -3,58 +3,102 @@ import axios from 'axios'
...
 
+  /**
+   * Builds the notification message to be sent,
+   * by composing the contents of the transformation event to readable
+   * message
+   * 
+   * @param event event to extract transformation results from 
+   * @returns message to be sent as notification
+   */
+  private buildMessage(event: TransformationEventInterface) {
```
georg-schwarz: Return type missing

notification/src/jsNotificationService.ts
```diff
@@ -3,58 +3,102 @@ import axios from 'axios'
...
+          Line: ${jobError.lineNumber}
+          Message: ${jobError.message}
+          Stacktrace: ${ jobError.stacktrace}`
+    } else {
+    /*======================================================
+     *  Build Message for succesfull transformation/pipline
+     *=======================================================*/ 
+      // Build Stats (Time measures for transformation execution)
+      const jobStats = event.jobResult.stats
+      let start = new Date(jobStats.startTimestamp)
```
georg-schwarz: Use `const` if value doesn't change, else use `let`

notification/src/jsNotificationService.ts
```diff
@@ -3,58 +3,102 @@ import axios from 'axios'
...
-        case 'FCM':
-            await this.handleFCM(notification as Firebase, message)
-            break
-        case 'SLACK':
-            await this.handleSlack(notification as Slack, message)
-            break
-        default:
-            throw new Error('Notification type not implemented.')
-        }
+    var message: string
```
georg-schwarz: use `let` instead

notification/src/models/notificationSummary.ts
```diff
@@ -0,0 +1,16 @@
+import { WebHookConfig, SlackConfig, FirebaseConfig } from './notificationConfig';
+
+export class NotificationSummary{
```
georg-schwarz: Do we need a class, or is an interface enough? (you can use interfaces and instate them with `= {attr1: xy, attrb: yz, ...}` instead of using a constructor).
Interfaces are typically used for pure data models without additional functionality, only for holding data.

notification/src/models/transformationEvent.ts
```diff
@@ -0,0 +1,32 @@
...
+        this.jobResult = jobResult
+    }
+
+    /**
+     * Checks if this event is a valid Transformation event,
+     * by checking if all field variables exist and are set.
+     *
+     * @returns     true, if param event is a TransformationEvent, else false
+     */
+    public isValidTransformationEvent(): boolean {
```
georg-schwarz: Same here. I would rather use an interface instead of a class and add a non-class related method (you can have method outside of classes in Typescript) that validates an event:
`export isValidTransformationEvent(event: TransformationEvent): boolean {...}`

transformation/package.json
```diff
@@ -1,51 +1,58 @@
...
+        "eslint": "^6.8.0",
+        "eslint-config-standard": "^14.1.1",
+        "eslint-plugin-import": "^2.20.2",
+        "eslint-plugin-node": "^11.1.0",
+        "eslint-plugin-promise": "^4.2.1",
+        "eslint-plugin-standard": "^4.0.1",
+        "jest": "^25.3.0",
+        "ts-jest": "^25.3.1",
+        "tslib": "^1.11.1",
+        "typescript": "^3.8.3"
```
georg-schwarz: Some version got dropped here... you can use `npm update` to insert latest versions automatically

transformation/src/handlers/storageHandler.ts
```diff
@@ -0,0 +1,123 @@
+import { Connection, ConnectionOptions, createConnection, getConnection } from 'typeorm';
+import { TransformationConfig } from '../interfaces/TransormationConfig';
+
+
+/**=============================================================================================================
+ * This class handles Requests to the Nofification Database 
+ * in order to store and get Notification Configurations.
+ *==============================================================================================================*/
+export class StorageHandler {
```
georg-schwarz: Refactor to make non-static

transformation/src/model/transformationEvent.ts
```diff
@@ -0,0 +1,32 @@
+import { TransformationEventInterface } from "../interfaces/transformationEventInterface"
+import JobResult from "../interfaces/jobResult/jobResult"
+
+
+/**
+ * Event send by the Transformation service upon transformation finish
+ */
+export class TransformationEvent implements TransformationEventInterface {
+    pipelineID: number      // ID of the pipeline
```
georg-schwarz: `pipelineId` instead of `ID`

transformation/src/transformationEndpoint.ts
```diff
@@ -16,10 +20,12 @@ export class TransformationEndpoint {
   keycloak?: Keycloak
 
   transformationService: TransformationService
+  
 
   constructor (transformationService: TransformationService, port: number, auth: boolean) {
```
georg-schwarz: Pass storageHandler and amqpHandler as constructor parameters and initialize outside. Let's us insert mocks for testing

notification/package.json
georg-schwarz: Alternatively just `npm update` to fix all version dumps

notification/src/notificationEndpoint.ts
```diff
@@ -16,10 +20,13 @@ export class NotificationEndpoint {
   keycloak?: Keycloak
 
   NotificationService: NotificationService
+  storageHandler: NotificationRepository
```
georg-schwarz, line 23: members always start with small letter (lower camel case)

transformation/src/handlers/amqpHandler.ts
```diff
@@ -0,0 +1,242 @@
...
+ *       A channel to notify the notification service that a transformation is done.
+ *       (see TransformationEventInterface for details of the Event)
+ * 
+ *      * Job Channel:
+ *       --------------
+ *       A Channel for the consumption of Transformation Jobs.
+ *       All Jobs for the transformation service will be consumed here.
+ *       (see JobEvent for details of the Event)
+ */
+export class AmqpHandler{
```
georg-schwarz: Make non-static and inject to endpoint via constructor

notification/src/interfaces/sandboxExecutor.ts
ke45xumo: Yes. The data ofJobresult of the tranformation (sandbox execution) may return undefined on failure. I imagined the scenario that the user wants only to be informed if the pipeline/transformation fails. He can set the condition to "data == undefined" for that purpose.

I may consider changing the data to null on failure in the (transformation-)sandbox executor, so the user can check for the condition "data ==null".
(Does this make a difference at all?)

notification/src/models/notificationSummary.ts
ke45xumo: Interface for this entity should be enough. I  will change this today

notification/notification-db/Dockerfile
ke45xumo: You are right. I will delete this

notification/src/interfaces/sandboxExecutor.ts
georg-schwarz: Sounds like a good decision! Make sure to document it somewhere

georg-schwarz: Makes sense! Make sure to document it somewhere ;)

notification/src/handlers/amqpHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+     */
+    public async connect(retries: number, backoff: number) {
+        const rabbit_url = process.env.AMQP_SERVICE_HOST;
+        const rabbit_usr = process.env.AMQP_SERVICE_USER;
+        const rabbit_password = process.env.AMQP_SERVICE_PWD;
+        const rabit_amqp_url = 'amqp://' + rabbit_usr + ':' + rabbit_password + '@' + rabbit_url;
+
+        console.log("URL: " + rabit_amqp_url)
+
+        var established: boolean = false    // amqp service connection result
```
georg-schwarz, line 50: `let`

notification/src/handlers/amqpHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+        const handler: AmqpHandler = this   // for ability to access methods and members in callback
+        
+        for (let i = 0; i < retries; i++) {
+            await this.backOff(backoff)
+            await connect(rabit_amqp_url, function (error0: any, connection: Connection) {
```
georg-schwarz: if you replace `function (error0, connection) { ... }` with `(error0, connection) => { ... }` you can still access the class members and functions (here AmqpHandler) with `this` and don't need the self-reference as variable beforehand ;)

notification/src/handlers/amqpHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+     * @param backOff   Period to wait in seconds
+     */
+    private backOff(backOff: number): Promise<void> {
+        return new Promise(resolve => setTimeout(resolve, backOff * 1000));
+    }
+
+    private initChannel(connection: any) {
+        const handler: AmqpHandler = this   // for ability to access methods and members in callback
+
+        connection.createChannel(function (error1: Error, channel: any) {
```
georg-schwarz: Same here

notification/src/handlers/amqpHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+        Pipeline Name: "${transformationEvent.pipelineName}`)
+        
+        const isValid = this.isValidTransformationEvent(transformationEvent)
+
+        if (!isValid) {
+            console.error('Message received is not an Transformation Event')
+            return false
+        }
+
+        let handler = new StorageHandler()
```
georg-schwarz: If the variable doesn't change, please use `const`

georg-schwarz: We should have `storageHandler` as member of `AmqpHandler` and initialize it via constructor parameter.

notification/src/handlers/amqpHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+
+        let configs = handler.getConfigsForPipeline(transformationEvent.pipelineId)
+        let executor = new VM2SandboxExecutor()
+        let notificationService = new JSNotificationService(executor)
```
georg-schwarz: Also inject those via constructor

notification/src/handlers/amqpHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+
+
+        configs.then(config => {
```
georg-schwarz, line 142: Use `await` syntax instead. This means the method will change to `handleEvent(msg: *type*): Promise<void>`

notification/src/handlers/storageHandler.ts
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /**
+     * Gets all WebHook Configs from the database for a specific pipeline id
+     *
+     * @param pipelineId    Pipeline ID to get the WebHook Configs for
+     */
+    public async getWebHookConfigs(pipelineId: number): Promise<WebHookConfig[]> {
+        console.log(`Getting WebHook Configs with pipelineId ${pipelineId} from Database`)
+        var webHookConfigs: WebHookConfig[] = []
```
georg-schwarz: `let`

notification/src/handlers/storageHandler.ts
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /**
+     * Gets all Firebase Configs from the database for a specific pipeline id
+     *
+     * @param pipelineId    Pipeline ID to get the Firebase Configs for
+     */
+    public async getFirebaseConfigs(pipelineId: number): Promise<FirebaseConfig[]> {
+        console.log(`Getting Firebase Configs with pipelineId ${pipelineId} from Database`)
+        var firebaseConfigs: FirebaseConfig[] = []
```
georg-schwarz: `let`

notification/src/handlers/storageHandler.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    /**
+     * Persists a webhook config (provided by argument) to the config database
+     *
+     * @param webhookConfig    webhook config to persist
+     * @returns Promise, containing the stored webhook config
+     */
+    public saveWebhookConfig(webhookConfig: WebHookConfig): Promise<WebHookConfig> {
+        const postRepository = getConnection().getRepository(WebHookConfig)
+        webhookConfig = postRepository.create(webhookConfig)
```
georg-schwarz: Use other variable instead, otherwise might introduce unexpeted side-effects at caller side

transformation/src/handlers/amqpHandler.ts
```diff
@@ -0,0 +1,246 @@
...
+     */
+    public async connect(retries: number, backoff: number) {
+        const rabbit_url = process.env.AMQP_SERVICE_HOST;
+        const rabbit_usr = process.env.AMQP_SERVICE_USER;
+        const rabbit_password = process.env.AMQP_SERVICE_PWD;
+        const rabit_amqp_url = 'amqp://' + rabbit_usr + ':' + rabbit_password + '@' + rabbit_url;
+
+        console.log("URL: " + rabit_amqp_url)
+
+        var established: boolean = false        // indicator if the connection has been established
```
georg-schwarz, line 67: `let`

transformation/src/handlers/storageHandler.ts
```diff
@@ -0,0 +1,131 @@
+import { Connection, ConnectionOptions, createConnection, getConnection } from 'typeorm';
+import { TransformationConfig } from '../models/TransormationConfig';
+
+
+/**
+ * This class handles Requests to the Nofification Database 
+ * in order to store and get Notification Configurations.
+ */
+export class StorageHandler {
```
georg-schwarz: Refactor according to insights of notification service

transformation/src/transformationEndpoint.ts
```diff
@@ -57,7 +72,76 @@ export class TransformationEndpoint {
...
 
+  /**===========================================================================
+    * Handles a request to save a TransformationConfig
+    * This is done by checking the validity of the config and then save
+    * it to the database on success
+    *============================================================================*/
+  handleConfigRequest = async (req: Request, res: Response): Promise<void> => {
+    console.log(`Received Transformation config from Host ${req.connection.remoteAddress}`)
+
+    var transformationConfig = req.body as TransformationConfig
```
georg-schwarz: `const`

transformation/src/transformationEndpoint.ts
```diff
@@ -57,7 +72,76 @@ export class TransformationEndpoint {
...
+    console.log(`Received request for configs with pipeline id ${pipelineID} from Host ${req.connection.remoteAddress}`)
+
+    if (!pipelineID) {
+      console.error("Request for config: ID not set")
+      res.status(400).send('Pipeline ID is not set.')
+      return
+    }
+
+    // Get configs from database
+    let configs = this.storageHandler.getTransformationConfigs(pipelineID)
```
georg-schwarz: `const`

notification/src/handlers/storageHandler.ts
```diff
@@ -25,6 +29,27 @@ export class StorageHandler implements NotificationRepository {
...
+     * and initiliazing a repository for the notification config
+     * 
+     * @param retries:  Number of retries to connect to the database
+     * @param backoff:  Time in seconds to backoff before next connection retry
+     */
+    public init(retries: number, backoff: number): void {
+        console.log('Initializing storageHandler.')
+        const handler :StorageHandler = this
+        
+        this.initConnection(retries, backoff).then(connection => {
```
georg-schwarz: Pls use async/await syntax

notification/src/handlers/storageHandler.ts
```diff
@@ -38,7 +63,7 @@ export class StorageHandler implements NotificationRepository {
      *
      * @returns     a Promise, containing either a Connection on success or null on failure
      */
-    public async initConnection(retries: number, backoff: number): Promise<void> {
+    private async initConnection(retries: number, backoff: number): Promise<Connection|null> {
```
georg-schwarz, line 78: Wouldn"t expose the connection to outside of class. If something goes wrong, reject promise instead of returning null

notification/src/handlers/storageHandler.ts
```diff
@@ -83,18 +119,37 @@ export class StorageHandler implements NotificationRepository {
...
+    /**
+     * Deletes all configs in the database referring to given pipeline id.
+     * If one of the deletions fails all changes will be rolled back. 
+     * 
+     * @param pipelineId Id of the pipeline to delete the configs for
+     */
+    public deleteConfigsForPipelineID(pipelineId: number): void {
+        console.log(`Deleting all configs for pipeline id "${pipelineId}"`)  
+        
+        let condition = { "pipelineId": pipelineId }
```
georg-schwarz, line 157: Use `const` for non-changing variables

notification/src/handlers/storageHandler.ts
```diff
@@ -160,10 +214,9 @@ export class StorageHandler implements NotificationRepository {
      * @returns Promise, containing the stored slack config
      */
     public saveSlackConfig(slackConfig: SlackConfig): Promise<SlackConfig>  {
-        const postRepository = getConnection().getRepository(SlackConfig)
-        slackConfig = postRepository.create(slackConfig)
+        slackConfig = this.slackRepository.create(slackConfig)
```
georg-schwarz: Store as new variable instead of overwriting parameter

notification/src/handlers/storageHandler.ts
```diff
@@ -173,19 +226,74 @@ export class StorageHandler implements NotificationRepository {
      * @returns Promise, containing the stored firebase config
      */
     public saveFirebaseConfig(firebaseConfig: FirebaseConfig): Promise<FirebaseConfig>  {
-        const postRepository = getConnection().getRepository(FirebaseConfig)
-        firebaseConfig = postRepository.create(firebaseConfig)
+        firebaseConfig = this.firebaseRepository.create(firebaseConfig)
```
georg-schwarz: Store as new variable instead of overwriting parameter

notification/src/interfaces/notificationRepository.ts
```diff
@@ -1,13 +1,25 @@
 import { FirebaseConfig, SlackConfig, WebHookConfig } from "../models/notificationConfig";
 import { NotificationSummary } from "./notificationSummary";
+import { Connection, DeleteResult, UpdateResult } from 'typeorm';
 
 export interface NotificationRepository {
-  initConnection(retries: number, backoff: number): Promise<void>
+  init(retries: number, backoff: number):void
```
georg-schwarz, line 6: Should return promise (await all the stuff that has to be inizialized)

notification/Dockerfile
```diff
@@ -3,6 +3,7 @@ FROM node:lts-alpine as builder
 
 WORKDIR /build
 COPY ./src ./src
+COPY ./tests ./tests
```
georg-schwarz, line 6: unit tests in js typically are in src directory and follow the pattern `{fileName}.test.js`

notification/jest.config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  preset: 'ts-jest',
-  testEnvironment: 'node',
-  roots: ['./src']
-}
+    preset: 'ts-jest',
+    testEnvironment: 'node',
+    roots: ['./tests']
```
georg-schwarz, line 4: unit tests in js typically are in src directory and follow the pattern `{fileName}.test.js`... so here `./src` would be the right value

notification/package.json
```diff
@@ -2,10 +2,10 @@
     "name": "open-data-service-notification",
     "version": "1.0.0",
     "description": "TBD",
-    "main": "dist/index.js",
+    "main": "dist/src/index.js",
```
georg-schwarz, line 5: Only `dist/index.js` required if tests are under `src`

notification/package.json
```diff
@@ -2,10 +2,10 @@
...
     "scripts": {
         "start": "npm run transpile && npm run start:transpiled",
-        "start:transpiled": "node dist/index.js",
+        "start:transpiled": "node dist/src/index.js",
```
georg-schwarz, line 8: Only `dist/index.js` required if tests are under `src`

notification/package.json
```diff
@@ -21,6 +21,7 @@
         "keycloak-connect": "^7.0.0",
         "pg": "^8.2.1",
         "reflect-metadata": "^0.1.13",
+        "sinon": "^9.0.2",
```
georg-schwarz: Please use `jest` for mocking and remove `sinon`

notification/package.json
ke45xumo: This was mandatory because with the new "tests" folder the compiler output will create two folders [ "tests", "src" ]

notification/Dockerfile
ke45xumo: Are you sure about that? 
I did that because the folder tests was introduced beforehand in tsconfig.json.
image
I thought it looks more  structured because you can differentiate tests from the productive system more clearly (especially when you have more tests).

Another benefit would be that you can only copy the binaries in the src folder (dist/src) from the intermediate container

notification/src/handlers/storageHandler.ts
ke45xumo: I think the connection is always exposed to the application.
You can always get the connection from the library via **getConnection()** function.

image

georg-schwarz: The issue with exposing the Connection is that we cannot wrap it with an implementation-neutral interface... Maybe we want to add another implementation with another library / database in the future.. That's why we shouldn't expose those implementation details ;-)

notification/Dockerfile
georg-schwarz: So, these structural changes have to be discussed with the other developers of the ODS. If we want to make those changes, we should do it on every service then. However, this should *not* be part of this pull request and discussed separatly... the PR is already huge enough

transformation/src/models/PipelineMetaData.ts
```diff
@@ -0,0 +1,40 @@
+import { Entity, PrimaryGeneratedColumn, Column, PrimaryColumn, JoinColumn, OneToOne } from 'typeorm';
+
+// psql -U transformation_usr -c "insert into pipeline_meta_data(\"displayName\", \"description\", \"author\", \"license\") values('a', 'b', 'c', 'd');INSERT into transformation_config (\"func\", \"datasourceId\", \"metadataId\") values ('aaaa',2 ,1)";
```
georg-schwarz: Can be removed?

transformation/src/models/TransformationConfig.ts
```diff
@@ -0,0 +1,23 @@
...
+    @Column()
+    datasourceId!: number;
+    
+    @OneToOne(type => PipelineMetaData, { eager: true, cascade: true, onDelete: "CASCADE" })
+    @JoinColumn()
+    metadata!: PipelineMetaData;
+
+    @OneToOne(type => TransformationData, { eager: true, cascade: true, onDelete: "CASCADE" })
+    @JoinColumn()
+    transformation!: TransformationData
```
georg-schwarz: Why as nested object and not as out-commented above as member `func`?

transformation/src/interfaces/jobResult/jobEvent.ts
```diff
@@ -0,0 +1,17 @@
...
+ * a channel.
+ * 
+ * It contains
+ * 
+ * @field pipelineID       pipeline ID of the request
+ * @field func             function to apply to the data (-> transformation) 
+ * @field data             data to be transformed
+ */
+export default interface JobEvent {
+    pipelineID: number      // pipeline ID of the request
```
georg-schwarz: pls use lower camel case `pipelineId`

transformation/src/handlers/amqpHandler.ts
```diff
@@ -0,0 +1,250 @@
...
+                return
+            }
+
+            handler.notificationChannel = channel
+
+            handler.notificationChannel.assertQueue(handler.notifQueueName, {
+                durable: false,
+            });
+
+            // // Consume from Channel
```
georg-schwarz: Deprecated?

transformation/src/handlers/storageHandler.ts
```diff
@@ -0,0 +1,307 @@
...
+
+
+/**
+ * This class handles Requests to the Nofification Database 
+ * in order to store and get Notification Configurations.
+ */
+export class StorageHandler implements TransformationRepository{
+
+    private configRepository!: Repository<TransformationConfig>
+    private dbConnection!: Connection | null
```
georg-schwarz, line 19: using a member with `!` but still allowing `null` seems weird to me

transformation/src/interfaces/jobResult/jobEvent.ts
```diff
@@ -0,0 +1,17 @@
...
+
+    func: string;           // function to apply to the data (-> transformation)
+
+    data: object;           // data to be transformed
```
georg-schwarz: Why do we need that? What is a JobEvent? 

Is it the communication between adapter and transformation? 
Then we do have a DataImportFinishedEvent with `datasourceId` as member, maybe containing the `data` if modeled as fat event. 

If it's for communication between UI and transformation for the preview of transformations, then use the REST API.

notification/src/notificationEndpoint.ts
```diff
@@ -3,11 +3,15 @@ import express, { Application, Request, Response } from 'express'
 import session, { MemoryStore } from 'express-session'
 import cors from 'cors'
 import Keycloak from 'keycloak-connect'
-import NotificationService from './interfaces/notificationService'
- 
+import NotificationService from './interfaces/notificationService';
+import "reflect-metadata"
```
georg-schwarz, line 7: Can be removed because imported in index?

transformation/src/models/TransformationConfig.ts
ke45xumo: I went with the definition of the pipeline config by the UI.

transformation/src/handlers/storageHandler.ts
ke45xumo: I had to assign null due to result of a `createConnection()` - Call. The `!` is a result of not initializing it in the constructor.
Unfortunately I have no clue to achieve this in another way.

transformation/src/interfaces/jobResult/jobEvent.ts
ke45xumo: A jobEvent is the Event that is published by the adapter and consumed by the transformation service. 
We did not agreed yet wether to use a fat or a thin event for that, so I kept both possibilities in the Event definition.

mathiaszinnen: Let's use the thin event for now since there are fewer changes necessary. After merging the PR we can decide whether we want to use fat events and refactor if needed.

docker-compose.yml
```diff
@@ -103,6 +103,11 @@ services:
       - "traefik.http.routers.to-scheduler.middlewares=scheduler-stripprefix@docker"
       - "traefik.http.middlewares.scheduler-stripprefix.stripprefix.prefixes=/api/scheduler"
 
+  scheduler-db:
```
mathiaszinnen: Why did you add persistence to the scheduler? I think you should remove the scheduler's persistence since we will keep the state in memory. Or is there something I am missing here?

ke45xumo: I added it by mistake. I will remove it with the next commit

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/model/AdapterConfig.java
```diff
@@ -8,6 +8,8 @@
 
 public class AdapterConfig {
 
+    private long id;
```
georg-schwarz, line 11: Where comes the id from? For what do we need it? AdapterConfig = ExecutionConfig, DatasourceConfig = stored config that leads to AdapterConfig for execution

notification/src/handlers/amqpHandler.ts
```diff
@@ -111,11 +132,9 @@ export class AmqpHandler{
             return false
         }
 
-        const handler = new StorageHandler()
 
-        const configs = handler.getConfigsForPipeline(transformationEvent.pipelineId)
-        const executor = new VM2SandboxExecutor()
-        const notificationService = new JSNotificationService(executor)
+        const configs = this.storageHandler.getConfigsForPipeline(transformationEvent.pipelineId)
+        const notificationService = new JSNotificationService(this.executor)
```
georg-schwarz: Should be injected from outside?

storage-mq/README.md
```diff
@@ -0,0 +1,105 @@
...
+TBD
+
+## API
+| Endpoint  | Method  | Request Body  | Response Body | Description |
+|---|---|---|---|---|
+| *base_url*/ | GET | - | text | Get health status |
+| *base_url*/version | GET | - | text | Get service version |
+| *base_url*/webhook | POST | webhookRequest | - | Trigger webhook |
+| *base_url*/slack | POST | slackRequest | - | Trigger slack notification |
+| *base_url*/fcm | POST | firebaseRequest | - | Trigger firebase notification |
```
georg-schwarz, line 36: Remove that copied stuff ;)

storage-mq/keycloak.json
```diff
@@ -0,0 +1,8 @@
+{
```
georg-schwarz, line 1: file can be removed

storage-mq/package.json
```diff
@@ -0,0 +1,50 @@
...
+        "transpile": "tsc",
+        "lint": "./node_modules/.bin/eslint src --ext .ts,.js --fix",
+        "test": "jest"
+    },
+    "dependencies": {
+        "axios": "^0.19.2",
+        "cors": "^2.8.5",
+        "express": "^4.17.1",
+        "express-session": "^1.17.1",
+        "keycloak-connect": "^10.0.2"
```
georg-schwarz: Remove keycloak

storage-mq/package.json
```diff
@@ -0,0 +1,50 @@
...
+    },
+    "devDependencies": {
+        "@types/amqplib": "^0.5.13",
+        "@types/deep-equal": "^1.0.1",
```
georg-schwarz, line 28: Some types are unnecessary here (deep-equal)

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/model/AdapterConfig.java
ke45xumo: I think i only messed up with the naming in the event (it should be adapterconfigId instead of datasourceId).

If i got that correct, then the pipeline config is referring to the adapter config id.

I only need the adapter config id to search corresponding pipeline configs in the transformation database.
Everything keeps on being the same, except that the transformation service knows which pipeline configs 
to extract from its database in order to get the transformations functions.

The datasource id will be still passed so the transformation service knows where to get the data from (when using thin event).

So the execution plan for thin event would be something like this:
1. scheduler triggers adapter import
2. adapter sends datasource link to data and corresponding adapter config id to transformation service (via event)
3. transformation gets data from send data source link
3. transformation service searches all pipeline configs that refer to the adapter cofnig id
4. transformation executes transformation functions of all found pipelines on the data from datasourcelink
5. transformation sends event to notification service (which handles notifications...)
5. transformation stores all transformation results to the new storage service (via Event)
6. the data can be accessed on storage service (via HTTP-GET)

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/model/AdapterEvent.java
```diff
@@ -0,0 +1,94 @@
+package org.jvalue.ods.adapterservice.adapter.model;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import java.io.Serializable;
+
+/**
+ * This Event will be sent to the transformation service after data import execution.
```
mathiaszinnen: better say `will be published to the import queue`. The adapter should be decoupled from the consumer of the events it puts on the queue

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/model/AdapterEvent.java
```diff
@@ -0,0 +1,94 @@
...
+
+  /**
+   * Default constructor
+   * @param datasourceId Adapter config identifier (referenced by pipeline config)
+   * @param data imported data (by adapter import)
+   * @param dataLocation location of the data on adapter service (--> endpoint)
+   */
+    public AdapterEvent(long datasourceId, String data, String dataLocation) {
+      this.datasourceId = datasourceId;
+      this.data = data;
```
mathiaszinnen, line 24: please use a thin event first (i.e. do not include the data itself). This reduces the number of changes necessary to make this pull request work. Afterward, we can change that, but in a different pull request.

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/model/AdapterConfig.java
mathiaszinnen: > I think i only messed up with the naming in the event (it should be adapterconfigId instead of datasourceId).
> If i got that correct, then the pipeline config is referring to the adapter config id.

Maybe our structure and naming are a bit misleading. Conceptually, an adapterConfig refers to a stateless data import and does not have an id. 
A datasourceConfig, on the other hand, refers to the datasource itself for which multiple data imports (using an adapter config) can be performed.  
The pipelines, therefore, do not have a reference to adapterConfigs but to data sources, which are persisted in the adapter service as datasourceConfig. 
That means your current naming in the event (datasourceId) is correct. In an event about imported data, we need the information to which data source (NOT adapterConfig = data import) the data belongs. 
In another(!) pull request, we should probably rename adapterConfig to dataImportConfig or refactor the adapter structure altogether to make it less confusing. 
Apart from that, I think your execution plan looks good and it is probably just a misconception about the naming of the concepts. 
If what I am writing does not make sense to you, we could talk about that in meet or something, just drop me a message in gitter if that would help you.

docker-compose.yml
```diff
@@ -111,8 +111,8 @@ services:
       DATABASE_HOST: storage-db
       DATABASE_PORT: 5432
       DATABASE_NAME: ods
-      DATABASE_USER: postgrest
-      DATABASE_PW: postgres-wrapper-pw
+      DATABASE_USER: ods_admin
```
mathiaszinnen: I think if you change the credentials here, you have to update the liquibase properties (`/storage/liquibase/changelog.xml`) accordingly.

notification/src/handlers/amqpHandler.ts
```diff
@@ -0,0 +1,176 @@
...
+        const rabit_amqp_url = 'amqp://' + rabbit_usr + ':' + rabbit_password + '@' + rabbit_url;
+
+        console.log("URL: " + rabit_amqp_url)
+
+        var established: boolean = false    // amqp service connection result
+        const handler: AmqpHandler = this   // for ability to access methods and members in callback
+        let errMsg: string = ''             // Error Message to be shown after final retry
+        
+        for (let i = 1; i <= retries; i++) {
+            await this.backOff(backoff)
```
mathiaszinnen, line 55: I think we should move this to the end of the method. Otherwise, we always wait, even when there is no error.

notification/src/handlers/amqpHandler.ts
```diff
@@ -0,0 +1,176 @@
...
+     * Waits for a specific time period.
+     *
+     * @param backOff   Period to wait in seconds
+     */
+    private backOff(backOff: number): Promise<void> {
+        return new Promise(resolve => setTimeout(resolve, backOff * 1000));
+    }
+
+    private initChannel(connection: any) {
+        console.log(`Initializing Transformation Channel "${this.notifQueueName}"`)
```
mathiaszinnen, line 91: Shouldn't it be Notification Channel?

notification/src/interfaces/sandboxExecutor.ts
```diff
@@ -1,3 +1,3 @@
 export default interface SandboxExecutor {
-  evaluate(expression: string, data: object): boolean;
+  evaluate(expression: string, data?: object): boolean;
```
mathiaszinnen, line 2: Why is data optional? I think we need data to evaluate the condition

notification/src/interfaces/transformationResults/transformationEvent.ts
```diff
@@ -0,0 +1,16 @@
...
+import JobResult from './jobResult';
+
+/**
+ * Event send by the Transformation service upon transformation finish
+ */
+export interface TransformationEvent {
+    pipelineId: number      // ID of the pipeline
+    pipelineName: string    // Name of the pipeline
+
+    dataLocation: string    // url (location) of the pipeline
```
mathiaszinnen, line 11: location of the data?

notification/src/jsNotificationService.ts
```diff
@@ -3,67 +3,113 @@ import axios from 'axios'
...
     }
 
+    return message
+  }
 
-  private async handleWebhook (webhook: Webhook): Promise<void> {
+  private async handleWebhook (webhook: WebHookConfig, message: string): Promise<void> {
     const callbackObject: WebhookCallback = {
-      location: webhook.dataLocation,
+      location: message,
```
mathiaszinnen: The `location` attribute should be renamed now that it contains a message (and not just a location).

notification/src/models/notificationConfig.ts
```diff
@@ -0,0 +1,65 @@
+import { Entity,  Column,  PrimaryGeneratedColumn } from 'typeorm'
+
+export enum CONFIG_TYPE{
+  WEBHOOK = "webhook",
+  SLACK = "slack",
+  FCM = "fcm"
+}
+
+
+export class NotificationConfig{
```
mathiaszinnen, line 10: Can two instances of notificationConfig have the same id because they are instances of different subtypes? I do not know if there is an easy way to fix this, but this seems inconsistent to me.

notification/src/interfaces/sandboxExecutor.ts
ke45xumo: This adds some functionality to the platform.

I thought of the use case that the user might have only notifications sent to another platform if the pipeline/transormation fails (this might make more sense with tightly scheduled jobs)

In this case he might set the condition to `data === null`

notification/src/models/notificationConfig.ts
ke45xumo: With the current database architecture this cannot be achieved and requires major refactoring on the service side to implement it in a proper way. A solution i can think of is that we can genereate a composed identifier that consists of the type and the autogenerated id (e.g. `SLACK_1` or `SLACK_1` cast as an number)

notification/src/notificationEndpoint.ts
ke45xumo: need it for type definitions in constructor parameter and member variable

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/model/AdapterEvent.java
ke45xumo: Currently the system is capable to handle both. 
If `data` exists (!=null) within the event it will use that. Else it will get the data from the datasource link `dataLocation`  (also in the event).

docker-compose.yml
ke45xumo: This is the old password (=equal to master branch credentials) .
I changed it because the liquibase service was failing (because of \r signs in the entrypoint.sh file) back then. 
I changed it back to the original values after the fix.

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/amqp/AmqpReceiver.java
```diff
@@ -0,0 +1,24 @@
+package org.jvalue.ods.adapterservice.adapter.amqp;
```
mathiaszinnen, line 1: I suppose you add that for communication between scheduler and adapter to trigger imports? 
If there is no reason why this absolutely needs to be included, please leave that out for now (since it is independent of the other changes) and do it later in another pull request.

adapter/src/main/java/org/jvalue/ods/adapterservice/config/RabbitConfiguration.java
```diff
@@ -4,12 +4,20 @@
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
+import static java.lang.System.getenv;
+
 @Configuration
 public class RabbitConfiguration {
     public static final String DATA_IMPORT_QUEUE = "data_import";
+    public static final String DATA_CONFIG_QUEUE = getenv("AMQP_ADAPTER_CONFIG_CHANNEL");
```
mathiaszinnen, line 12: If possible, please leave adapter/scheduler communication out of this pr

georg-schwarz:
Closed it! We broke it down into smaller digestable pieces

PR #114, opened 2020-07-02
Datasource trigger integration test

jenswaechtler:
The integration test for the datasource trigger endpoint. There is a new endpoint for the mock server that dynamically responds to a URI. The test is done once with free variables and once without free variables with an empty POST request.
@georg-schwarz @mathiaszinnen

### DIFF
--- a/adapter/integration-test/src/adapter-datasource.test.js
+++ b/adapter/integration-test/src/adapter-datasource.test.js
@@ -2,6 +2,9 @@ const request = require('supertest')
 const waitOn = require('wait-on')
 
 const URL = process.env.ADAPTER_API || 'http://localhost:9000/api/adapter'
+const MOCK_SERVER_PORT = process.env.MOCK_SERVER_PORT || 8081
+const MOCK_SERVER_HOST = process.env.MOCK_SERVER_HOST || 'localhost'
+const MOCK_SERVER_URL = 'http://' + MOCK_SERVER_HOST + ':' + MOCK_SERVER_PORT
 
 describe('Adapter Configuration', () => {
   console.log('Adapter-Service URL= ' + URL)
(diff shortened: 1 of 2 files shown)


jenswaechtler:
On my local machine the test runs without errors. Do you know what the problem could be?

adapter/integration-test/src/adapter-datasource.test.js
```diff
@@ -184,6 +187,52 @@ describe('Adapter Configuration', () => {
...
 })
 
+test('POST datasources/{id}/trigger', async () => {
+  const dynamicDatasourceResponse = await request(URL)
+    .post('/datasources')
+    .send(dynamicDatasourceConfig)
+  const dynDatasourceId = dynamicDatasourceResponse.body.id
+
+  const dynamicDataMetaData = await request(URL)
+    .post('/datasources/' + dynDatasourceId + '/trigger')
```
georg-schwarz: Little pointer to how you can concat strings very nicely in JavaScript:
```
`/datasources/${dynDatasourceId}/trigger`
```

adapter/integration-test/src/adapter-datasource.test.js
```diff
@@ -184,6 +187,52 @@ describe('Adapter Configuration', () => {
   })
 })
 
+test('POST datasources/{id}/trigger', async () => {
```
georg-schwarz: Does it make sense to split dynamic and non-dynamic into separate test cases?

georg-schwarz:
Integrationtest for adapter red, internal server error!

adapter/integration-test/src/adapter-datasource.test.js
jenswaechtler: Maybe, but actually it should be explicit for the trigger endpoint. I followed the other tests that test for the endpoint as well.

adapter/integration-test/src/adapter-datasource.test.js
jenswaechtler: Thanks for the suggestion. I changed it right away.

mathiaszinnen:
> On my local machine the test runs without errors. Do you know what the problem could be?

I'll look into it later

mathiaszinnen (commented):
The internal server error is probably caused by rabbitmq. I had the same problem before. RabbitMQ is started as a dependency but the trigger test runs before it is ready to take connections which causes the adapter to throw an exception. 
If you wait-on for rabbitmq in the datasource-integration test like it is done in the stateless test, the problem should be solved.

adapter/integration-test/src/adapter-datasource.test.js
jenswaechtler: The reasoning is that all data sources are treated as dynamic anyway. Either they have open parameters or not. But if you think that makes more sense, then it is fine with me.

mathiaszinnen (approved):

PR #119, opened 2020-07-03
Fix rabbitmq url for system-test

jenswaechtler:
At the moment it is not possible to run the system-test locally via `npm run test` without errors. This is because the URL for rabbitmq is undefined if the system-test is not started through Docker. This commit ensures that the locally running rabbitmq service is used in this case.


jenswaechtler:
notfiy @georg-schwarz @mathiaszinnen

mathiaszinnen (approved):
Very nice!

georg-schwarz (approved):

PR #121, opened 2020-07-09
Add example request for dynamic datasource

jenswaechtler:


mathiaszinnen (approved):

doc/example-requests/adapter.http
```diff
@@ -131,3 +131,45 @@ Content-Type: application/json
...
+
+### Create datasource with dynamic parameters
+POST {{baseURL}}/datasources  HTTP/1.1
+Content-Type: application/json
+
+{
+  "protocol": {
+    "type": "HTTP",
+    "parameters": {
+      "location": "https://www.pegelonline.wsv.de/webservices/rest-api/v2/stations/{station}/W/measurements.json?start=P1D",
```
georg-schwarz, line 147: Looks good!
Do we have to specify the parameters somewhere further? Like default value, etc.? Or is that an extra PR?

georg-schwarz:
Nice! Thanks @jenswaechtler for your PR! 
I like the size and the focus of your PRs!

Currently, we have a small merge freeze until #120 is merged... I hope to accept your PR on Monday ;)

doc/example-requests/adapter.http
jenswaechtler: That is coming in supplemental PR. The problem is that for default values I have to touch the model of the datasource. So I must probably touch a couple of files and it would get unnecessarily cluttered.

PR #129, opened 2020-07-14
Scheduler notification mq

mathiaszinnen:


scheduler/src/clients/amqp-client.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+  } else {
+    console.log('AmqpClient initialization aborted: Already initialized.')
+    return
+  }
+}
+
+async function connect(): Promise<amqp.Connection> {
+  try {
+    const connection = await amqp.connect(AMQP_URL)
```
georg-schwarz, line 33: Having a retry machanism would be nice here! (see storage-mq or notification)

scheduler/src/workflow-execution.test.ts
```diff
@@ -92,7 +95,8 @@ test('Should trigger notifications', async () => {
 
   await PipelineExecution.execute(datasourceConfig)
 
-  expect(mockExecuteNotification).toHaveBeenCalledTimes(1)
+  expect(mockInitAmqp).toHaveBeenCalledTimes(1)
```
georg-schwarz: Init should be called in index.ts (at least I would expect that)?

scheduler/src/workflow-execution.ts
```diff
@@ -89,21 +90,26 @@ async function executeStorage (args: { pipelineConfig: PipelineConfig, data: obj
...
     dataLocation: args.dataLocation,
     pipelineId: pipelineConfig.id,
     pipelineName: pipelineConfig.metadata.displayName,
     data: data,
     error: error
   }
 
-  await NotificationClient.executeNotification(notificationTrigger)
-  console.log(`Successfully delivered notification request to transformation-service for ${pipelineConfig.id}`)
+  const success = AmqpClient.publish(notificationTrigger)
```
georg-schwarz, line 116: await I think

docker-compose.yml
```diff
@@ -93,8 +93,10 @@ services:
       CONFIG_SERVICE_URL: http://core:8080
       ADAPTER_SERVICE_URL: http://adapter:8080
       TRANSFORMATION_SERVICE_URL: http://transformation:8080
-      NOTIFICATION_SERVICE_URL: http://notification:8080
       STORAGE_SERVICE_URL: http://storage:3000
+      AMQP_URL: amqp://rabbit_adm:R4bb!7_4DM_p4SS@rabbitmq:5672
+      AMQP_EXCHANGE: ods_global
+      AMQP_NOTIFICATION_EXECUTION_TOPIC: notification.execution.*
```
georg-schwarz: I think for publishing you will need the exact topic.
Additionally, I would call the topic according to what happened, like "pipeline.execution.success"

georg-schwarz (commented):
Looks good to me so far! Added some comments to a few smaller things

scheduler/src/workflow-execution.ts
mathiaszinnen: The amqplib we use only offers a sync publish method. I was surprised by that too but decided to keep the wrapper asynchronous as well. Do you think I should wrap it asynchronously or try to find a way to have the publication acked?

scheduler/src/clients/amqp-client.ts
mathiaszinnen: good point

scheduler/src/workflow-execution.ts
georg-schwarz: Getting an ACK is good, even better if it would work async... but just make it work for the time being, there is still time for those optimizations afterwards

scheduler/integration-test/src/scheduler.test.js
```diff
@@ -99,6 +104,19 @@ describe('Scheduler', () => {
   })
 })
 
+async function receiveAmqp (url, exchange, topic, queue) {
+  const connection = await amqp.connect(url)
+  const channel = await connection.createChannel()
+  await channel.assertExchange(exchange, 'topic', {durable: true})
+  const q = await channel.assertQueue(queue)
+  await channel.bindQueue(q.queue, exchange, topic)
+  await channel.consume(q.queue, consumeEvent)
```
georg-schwarz, line 128: You can look into PR #128 how I solved the AMQP connection in the integration tests....
Important is to also close the connection, otherwise Jest won't shutdown ;)

scheduler/src/index.ts
```diff
@@ -91,6 +92,7 @@ process.on('unhandledRejection', (reason: any, promise: Promise<any>) => {
 // eslint-disable-next-line @typescript-eslint/no-floating-promises
 initPipelineConfigSync()
 initJobs()
+AmqpClient.init(3, 2)
```
georg-schwarz: Make total wait time ~60s

georg-schwarz (approved):

PR #130, opened 2020-07-19
Define default runtime parameters

jenswaechtler:
I modified the Http Importer to handle default values. Since default values are not needed in every case, I have built in the ability to differentiate between available and required parameters.
What is your opinion on this? @georg-schwarz @mathiaszinnen

### DIFF
--- a/adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/importer/HttpImporter.java
+++ b/adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/importer/HttpImporter.java
@@ -1,5 +1,6 @@
 package org.jvalue.ods.adapterservice.adapter.importer;
 
+import org.jvalue.ods.adapterservice.datasource.model.RuntimeParameters;
 import org.springframework.web.client.RestTemplate;
 
 import java.net.URI;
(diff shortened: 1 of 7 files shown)


adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/importer/HttpImporter.java
```diff
@@ -49,6 +53,10 @@ protected void validateParameters(Map<String, Object> inputParameters) {
     return parameters;
   }
 
+  @Override
+  protected List<ImporterParameterDescription> getRequiredParameters() {
```
georg-schwarz, line 57: I think we can do it that way!
Another possibility would be to add a boolean field to `ImpoerterParameterDescription` that states if the parameter is required or not. Actually I think that is a good idea, then we could also default other parameters like encoding with a value. But probably better subject of another PR.

jenswaechtler: Okay, good point. I will put it in an upcoming PR.

georg-schwarz (approved):

PR #131, opened 2020-07-21
Required parameters for importer

jenswaechtler:
I changed the validation. Now you can define with a constructor if an import parameter is needed or not. Default behavior is required. Additionally I changed that you cannot use import parameters which are not defined for the specific importer.


georg-schwarz (approved):
Nice!

PR #155, opened 2020-08-05
Update typescript target to es2018

sonallux:
- Update typescript `target` and `lib` to es2018 as this is the latest supported version by NodeJS LTS 
- Fix typos
- Add missing `.eslintrc` for notification service. Now all services have the same eslint configuration.

### DIFF
--- a/notification/.eslintrc
+++ b/notification/.eslintrc
@@ -1,0 +1,25 @@
+{
+  "env": {
+    "browser": true,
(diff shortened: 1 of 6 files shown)


mathiaszinnen (approved):
Looking good, thanks!

georg-schwarz (approved):
Nice! Didn't have that on the radar at all

Issue #156, opened 2020-08-06
Validation in Adapter API


mathiaszinnen:
When creating/updating datasources in the `adapter/datasources` API, validation does not work as expected (e.g. posting a datasource with protocol type "LOL" does not trigger malformed request).
When trying to trigger an import for a malformed datasource with the `datasources/{id}/trigger` endpoint, all exceptions that occur while importing are mapped to Http 404 responses. 
We should refactor input validation for datasource management and exception handling for the trigger part.

sonallux:
This can be closed, as the two main points of this issue have been resolved in #231 and #235.

PR #159, opened 2020-08-06
Adapter amqp transformation

mathiaszinnen:


adapter/integration-test/src/adapter-stateless.test.js
```diff
@@ -8,23 +8,13 @@ const MOCK_SERVER_URL = 'http://' + MOCK_SERVER_HOST + ':' + MOCK_SERVER_PORT
...
-    }
-
     try {
       const pingUrl = URL + '/version'
-      console.log('Waiting for service with URL: ' + pingUrl)
-      await waitOn({ resources: [pingUrl], timeout: 50000 })
-      console.log('[online] Service with URL:  ' + pingUrl)
+      console.log('Starting adapter stateless test')
+      console.log(`Waiting for services: ${pingUrl} ${MOCK_SERVER_URL}, ${RABBIT_URL}`)
+      await waitOn({ resources: [MOCK_SERVER_URL, RABBIT_URL, pingUrl], timeout: 50000 })
```
georg-schwarz: use `log: true` for waitOn instead of loggin beforehand manually

transformation/src/index.ts
```diff
@@ -44,11 +45,14 @@ const server = app.listen(port, async () => {
     executionResultPublisher
   )
 
+  const pipelineConfigConsumer = new PipelineConfigConsumer(pipelineConfigManager)
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const pipelineExecutionEndpoint = new PipelineExecutionEndpoint(pipelineExecutor, app)
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const pipelineConfigEndpoint = new PipelineConfigEndpoint(pipelineConfigManager, app)
 
+  await pipelineConfigConsumer.connect(30, 2000)
```
georg-schwarz: use variables instead of magic numbers

adapter/integration-test/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "mock": "node src/mock.server.js",
     "pretest": "npm run mock &",
-    "test": "jest",
+    "test": "jest --runInBand",
```
georg-schwarz: Do we need this?

adapter/integration-test/src/adapter-stateless.test.js
mathiaszinnen: Nice, will do!

adapter/integration-test/package.json
mathiaszinnen: I changed that for local debugging and will change back. I also think we need faster ci runs

georg-schwarz: Allright :) yeah I'm also a bit concerned about the ci times, feels like way too much - but most of it results from docker artifact up- and download..

adapter/integration-test/src/adapter-datasource-trigger.test.js
```diff
@@ -0,0 +1,263 @@
...
+      .post(`/datasources/${datasourceId}/trigger`)
+      .send(runtimeParameters)
+
+    const id = dataMetaData.body.id
+    const data = await request(URL)
+      .get(`/data/${id}`)
+      .send()
+
+    const delResponse = await request(URL)
+      .delete('/datasources/')
```
georg-schwarz: Please only delete the created one, otherwise we won't come back to running the integration tests without `--runnInBands` flag at all

adapter/integration-test/src/adapter-datasource-trigger.test.js
```diff
@@ -0,0 +1,263 @@
...
+      .post('/datasources')
+      .send(brokenDatasourceConfig)
+
+    const datasourceId = datasourceResponse.body.id
+
+    const triggerResponse = await request(URL)
+      .post(`/datasources/${datasourceId}/trigger`)
+      .send()
+
+    expect(triggerResponse.status).toBeGreaterThan(300) // request should fail (no 2xx status)
```
georg-schwarz, line 162: why not make concrete? either 400 for bad request, or imho better 404 (because resource does not exist)

transformation/src/api/amqp/pipelineConfigConsumer.ts
```diff
@@ -0,0 +1,83 @@
+import * as AMQP from 'amqplib'
+import { PipelineConfigManager } from '../../pipeline-config/pipelineConfigManager'
+import pipelineConfigTriggerRequest from '../pipelineConfigTriggerRequest'
```
georg-schwarz, line 3: Use `env.ts` without default values for fail-fast on startup with wrong configuration

adapter/integration-test/src/adapter-datasource-trigger.test.js
mathiaszinnen: I chose to leave it open because the validation in the adapter service does not work now and needs refactoring. It currently returns 404 no matter what error occurs, see #156

adapter/integration-test/src/adapter-datasource-trigger.test.js
mathiaszinnen: yeah, good point

georg-schwarz (approved):
Nice, feels good to finish the migration! :)

PR #162, opened 2020-08-10
liquibase: Add retrying of update

sonallux:
### Issue
When running `docker-compose up -d` I often had the problem, that the execution of liquibase failed, because the postgres database was not yet fully setup. This would then lead to further problems as the `storage` and the `storage-mq` service can not access the database because their database users are created by liquibase.

### Fix
This PR fixes this issue by adding retry functionality to the liquibase docker container. If the liquibase update is failing, it will now be retried after a short sleep. The maximal number of connection retries and the sleep duration can be changed with the `CONNECTION_RETRIES` and `CONNECTION_BACKOFF_IN_SECONDS` environment variables.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -145,6 +145,10 @@ services:
       DATABASE_NAME: ods
       DATABASE_USER: ods_admin
       DATABASE_PW: ods_pw
+      CONNECTION_RETRIES: 5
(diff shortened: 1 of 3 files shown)


georg-schwarz (approved):
Nice, good job!

PR #172, opened 2020-08-28
UI: Remove usage of `as` operator

### LINKED ISSUE #171: Wrong usage of TypeScripts `as` operator
While developing #170 I noticed a lot of wrong usages of TypeScript's `as` operator. The `as` operator is a [type assertion](https://www.typescriptlang.org/docs/handbook/basic-types.html#type-assertions) to tell the typescript compiler that you know more about a value than the TypeScript compiler does. 
> Type assertions are a way to tell the compiler “trust me, I know what I’m doing.” A type assertion is like a type cast in other languages, but performs no special checking or restructuring of data. **It has no runtime impact, and is used purely by the compiler**. TypeScript assumes that you, the programmer, have performed any special checks that you need.

I have seen a lot of cases where a `string` should be transformed into a `number`. But this is actually not done during runtime:
```ts
// input is of type: string
const id = (input as unknown) as number
// doSomething is of type: (id: number): void
doSomething(id)
```
In many cases, this is not a problem as JavaScript is smart enough to transform the string into a number when it is actually needed. But it is an issue when checking the type of a value with JavaScript's `typeof` operator.

I propose to get rid of all those type assertions in all microservices. There is also a [typescript-eslint rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-assertions.md) to enforce this.

- [x] UI #172 
- [x] transformation service #181
- [x] notification service #182
- [x] scheduler service #185
- [x] storage-mq service #187

sonallux:
- Remove usage of TypeScript's `as` operator (See #171)
- Fix some eslint errors

### DIFF
--- a/ui/.eslintrc
+++ b/ui/.eslintrc
@@ -28,6 +28,7 @@
     "@typescript-eslint/indent": [2, 2],
     "@typescript-eslint/no-use-before-define": ["warn", { "functions": false, "classes": true }],
     "max-len": [2, 120, 4, { "ignoreUrls": true }],
-    "@typescript-eslint/explicit-function-return-type": [1, {"allowExpressions": true}]
+    "@typescript-eslint/explicit-function-return-type": [1, {"allowExpressions": true}],
+    "@typescript-eslint/consistent-type-assertions": [ "warn", { "assertionStyle": "never" }]
   }
 }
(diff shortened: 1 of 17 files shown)


andreas-bauer:
Thanks, PR looks good.

andreas-bauer (approved):

PR #173, opened 2020-08-31
UI: Fix DatasourceEdit and PipelineEdit

sonallux:
~PR #172 introduced a weird bug with `@State` or `@Watch` decorators that they are not working with a value of `undefined` for the `selectedDatasource` and `selectedPipeline` properties in the `VuexModule`. I did not find a simple way of fixing this bug. But I decided to fix this bug by removing the `selectedDatasource` or `selectedPipeline` properties from the `VuexModule` and directly using `datasourceRest.ts` and `pipelineRest.ts` because of three reasons:~
- ~This fixes the bug mentioned above~
- ~Vuex adds a lot of complexity to the UI code~
- ~Caching of the `selectedDatasource` or `selectedPipeline` properties in the VuexModule does not make much sense. It is only used when creating the `DatasourceEdit` or `PipelineEdit` component.~

Fix fixes an issue because I forgot to initialize the properties correctly...

### DIFF
--- a/ui/src/datasource/datasource.module.ts
+++ b/ui/src/datasource/datasource.module.ts
@@ -6,7 +6,7 @@ import * as RestService from './datasourceRest'
 export default class DatasourceModule extends VuexModule {
   private datasources: Datasource[] = []
 
-  private selectedDatasource?: Datasource
+  private selectedDatasource?: Datasource = undefined
 
   private isLoadingDatasources = true
 
--- a/ui/src/notification/NotificationOverview.vue
+++ b/ui/src/notification/NotificationOverview.vue
@@ -120,15 +120,14 @@ export default class PipelineNotifications extends Vue {
   private loadConfigbyPipelineIdAction!: (id: number) => void
 
   @Action('addNotification', notificationNameSpace)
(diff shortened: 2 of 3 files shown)


sonallux:
I have forced push my changes to rerun the tests, because the adapter integration test are failing randomly...

andreas-bauer:
I guess there is a general agreement that the Vuex store adds in some cases more complexity than it provides benefits to a component.

@jsone-studios do you plan to remove the other Vuex operations in further PRs?
Because mixing Vuex operations and a custom data loading mechanism in one component doesn't make this component less complex or easier to understand.

sonallux:
> I guess there is a general agreement that the Vuex store adds in some cases more complexity than it provides benefits to a component.
> 
> @jsone-studios do you plan to remove the other Vuex operations in further PRs?
> Because mixing Vuex operations and a custom data loading mechanism in one component doesn't make this component less complex or easier to understand.

Actually, I do not have plans to do that. But I found a much simpler way of fixing the bug. I just need to initialize the properties correctly...

andreas-bauer (approved):

Issue #177, opened 2020-08-31
Pipeline execution leads to transformation exception


mathiaszinnen:
Executing the example pipeline from the ui does not work. The transformation service logs show the following exception:

``` 
Possibly Unhandled Rejection at: Promise Promise {
transformation_1 | <rejected> TypeError: given data is no object or array
transformation_1 | at VM2SandboxExecutor.assertIsObjectOrArray (/app/dist/pipeline-execution/sandbox/vm2SandboxExecutor.js:14:19)
transformation_1 | at VM2SandboxExecutor.execute (/app/dist/pipeline-execution/sandbox/vm2SandboxExecutor.js:18:14)
transformation_1 | at /app/dist/pipeline-execution/pipelineExecutor.js:20:79
transformation_1 | at PipelineExecutor.executionTimeInMillis (/app/dist/pipeline-execution/pipelineExecutor.js:13:24)
transformation_1 | at PipelineExecutor.executeJob (/app/dist/pipeline-execution/pipelineExecutor.js:20:37)
transformation_1 | at PipelineConfigManager.triggerConfig (/app/dist/pipeline-config/pipelineConfigManager.js:59:50)
transformation_1 | at processTicksAndRejections (internal/process/task_queues.js:97:5)
transformation_1 | at async PipelineConfigConsumer.consumeEvent (/app/dist/api/amqp/pipelineConfigConsumer.js:36:21)
transformation_1 | } reason: TypeError: given data is no object or array
transformation_1 | at VM2SandboxExecutor.assertIsObjectOrArray (/app/dist/pipeline-execution/sandbox/vm2SandboxExecutor.js:14:19)
transformation_1 | at VM2SandboxExecutor.execute (/app/dist/pipeline-execution/sandbox/vm2SandboxExecutor.js:18:14)
transformation_1 | at /app/dist/pipeline-execution/pipelineExecutor.js:20:79
transformation_1 | at PipelineExecutor.executionTimeInMillis (/app/dist/pipeline-execution/pipelineExecutor.js:13:24)
transformation_1 | at PipelineExecutor.executeJob (/app/dist/pipeline-execution/pipelineExecutor.js:20:37)
transformation_1 | at PipelineConfigManager.triggerConfig (/app/dist/pipeline-config/pipelineConfigManager.js:59:50)
transformation_1 | at processTicksAndRejections (internal/process/task_queues.js:97:5)
[...]

PR #178, opened 2020-09-01
Data import fat event

### LINKED ISSUE #174: storage-mq crashes at startup
The storage-mq crashes at startup. The exception thrown hints at the durable argument for ods_global exchange.

mathiaszinnen:
closes #174 and #177

### DIFF
--- a/transformation/src/api/amqp/pipelineConfigTriggerRequestAmpq.ts
+++ b/transformation/src/api/amqp/pipelineConfigTriggerRequestAmpq.ts
@@ -1,16 +1,16 @@
-import { isObject, isNumber, isString, hasProperty } from '../validators'
+import { isObject, isNumber, isString, hasProperty } from '../../validators'
 
-export interface PipelineConfigTriggerRequest {
+export interface PipelineConfigTriggerRequestAmpq {
   datasourceId: number;
 
   data: string;
   dataLocation: string;
 }
 
-export class PipelineConfigTriggerRequestValidator {
+export class PipelineConfigTriggerRequestAmqpValidator {
   private errors: string[] = []
 
-  validate (requestBody: unknown): requestBody is PipelineConfigTriggerRequest {
+  validate (requestBody: unknown): requestBody is PipelineConfigTriggerRequestAmpq {
     this.errors = []
     if (!isObject(requestBody)) {
       this.errors.push('\'PipelineConfigTriggerRequest\' must be an object')
@@ -21,8 +21,8 @@ export class PipelineConfigTriggerRequestValidator {
     } else if (!isNumber(requestBody.datasourceId)) {
       this.errors.push('\'datasourceId\' must be a number')
     }
-    if (hasProperty(requestBody, 'data') && !isObject(requestBody.data)) {
-      this.errors.push('\'data\' must be an object or array')
+    if (hasProperty(requestBody, 'data') && !isString(requestBody.data)) {
+      this.errors.push('\'data\' must be a string')
     }
     if (hasProperty(requestBody, 'dataLocation') && !isString(requestBody.dataLocation)) {
       this.errors.push('\'dataLocation\' must be a string')
--- a/transformation/src/api/amqp/pipelineConfigTriggerRequestAmpq.ts
+++ b/transformation/src/api/amqp/pipelineConfigTriggerRequestAmpq.ts
@@ -2,9 +2,7 @@ import { isObject, isNumber, isString, hasProperty } from '../../validators'
 
 export interface PipelineConfigTriggerRequestAmpq {
   datasourceId: number;
-
   data: string;
-  dataLocation: string;
 }
 
 export class PipelineConfigTriggerRequestAmqpValidator {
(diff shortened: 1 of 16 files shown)


transformation/src/api/pipelineConfigTriggerRequest.ts
```diff
@@ -1,6 +1,6 @@
 export default interface PipelineConfigTriggerRequest {
   datasourceId: number;
 
-  data: object;
+  data: string;
```
sonallux: Why this change? I think it is much cleaner to have the `data` already present as an object and not as stringified JSON.

mathiaszinnen: The data attribute is transferred as string (i.e. the json object is enclosed by ''). Changing the type to string was the quickest way of handling the conversion problem. I think you are right that the object type is cleaner, however. I think about if there is a easy solution to achieve that. Otherwise, we could move that to another issue

mathiaszinnen: The cleanest solution is changing the adapter behavior such that the enclosing quotes are removed, but this is more complex than it seems (because of persistence in the adapter). I would propose to create another issue for that and use the quick and dirty solution here

sonallux: Ok, I see, But your change will definitely break the REST API endpoint `/api/transformation/trigger` because here the `data` is an object. Maybe we should create two `PipelineConfigTriggerRequest` interfaces, one for amqp and one for rest.

sonallux: And for the `AmqpPipelineConfigTriggerRequest` you can also remove the `dataLocation` property, as it is not used/supported there.

sonallux: I think both of your options are not ideal and not easy to implement. Therefore, I would leave your current PR as it is and send the data as string for now. But I would create a new `AmqpPipelineConfigTriggerRequest` interface that is used inside `pipelineConfigConsumer.ts`:
```ts
export default interface AmqpPipelineConfigTriggerRequest {
  datasourceId: number;
  data: string;
}
```
And the `/api/transformation/trigger` endpoint will still use the old `PipelineConfigTriggerRequest` interface.

I think this solution is also not ideal, because sending the data via RabbitMQ can lead to issue if the data is very large. We should definitely create an issue for this and find a better way of sending the data to the transformation service.

georg-schwarz: Here my two cents on it.

"Make it work. Make it pretty. Make it fast." - that's the order we should go for. 

"Make it work": for a quick fix it is ok to make that change - however make sure to not break other stuff.
Regarding the REST trigger endpoint: is it even used anymore? Maybe we can also remove it if we don't need it anymore... I think it only was used for the migration to amqp.

"Make it pretty": we will have to decide whether to send the data as strings or as object. I have a hunch that the Java annotation magic limits us here (store as string, send as object) or at least makes it more complicated? 

"Make it fast": That's the point with what to send by events. The data itself as fat event, an URL where to fetch or even a reference to a database table as thin event. 
I believe sending fat events is the easiest thing but not the best performing. Making it efficient should be done as soon as we have a use case that brings up that as a bottleneck. IMHO no need to optimize where currently not needed - although it also inches me every time I see it ;)

transformation/src/api/amqp/pipelineConfigTriggerRequestAmpq.ts
```diff
@@ -0,0 +1,39 @@
+import { isObject, isNumber, isString, hasProperty } from '../../validators'
+
+export interface PipelineConfigTriggerRequestAmpq {
+  datasourceId: number;
+
+  data: string;
+  dataLocation: string;
```
sonallux:
```suggestion
```
This interface should match exactly the adapter's `DatasourceImportedEvent.java` class, which contains the structure of the event during creation.

sonallux (changes requested):

transformation/src/api/pipelineConfigTriggerRequest.ts
mathiaszinnen: > Regarding the REST trigger endpoint: is it even used anymore? Maybe we can also remove it if we don't need it anymore... I think it only was used for the migration to amqp.

I will create an issue to decide whether we want to keep it for convenience or remove it for simplicity

transformation/src/api/amqp/pipelineConfigTriggerRequestAmpq.ts
mathiaszinnen: The interface was created this way to enable both thin and fat events. Personally, I agree with you that we should keep the model as simple and possible and add extra fields only if needed, but I would like to hear @georg-schwarz opinion before changing this since we have been discussing the issue before.

sonallux: I missed mentioning in my comment, that the `dataLocation` property is also not used in the amqp `pipelineConfigConsumer.ts`. Therefore I would remove this property for now and if we want to support thin events, we can change it in another PR.

georg-schwarz: Well, there is no right and wrong here. Currently I tend to make it as easy as possible and agree on one. We can optimize later when we have a use case for it :)

Premature optimization is the mother of all evil they say ;)

But not sure if that is the right choice ...

mathiaszinnen: I happily removed it :) It is still present in the pipelineExecutionRequest, however. I did not remove it there because I want to be sure not to introduce another error (that is unrelated to the purpose of this pr). I create another issue for that.

mathiaszinnen:
Should be ready for merge now. Thank you @jsone-studios for your extensive feedback!

georg-schwarz (approved):

PR #187, opened 2020-09-08
Storage-mq: Remove usage of `as` operator

sonallux:
- Remove usage of `as` operator (See #171)
- Some code cleanup

### DIFF
--- a/adapter/integration-test/src/adapter-datasource-trigger.test.js
+++ b/adapter/integration-test/src/adapter-datasource-trigger.test.js
@@ -17,7 +17,7 @@ const EXECUTION_TOPIC = process.env.AMQP_IMPORT_TOPIC
 const EXECUTION_SUCCESS_TOPIC = process.env.AMQP_IMPORT_SUCCESS_TOPIC
 const EXECUTION_FAILED_TOPIC = process.env.AMQP_IMPORT_FAILED_TOPIC
 
-const STARTUP_DELAY = 2000
+const STARTUP_DELAY = 5000
 
 let amqpConnection
 const publishedEvents = new Map() // routing key -> received msgs []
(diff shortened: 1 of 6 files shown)


storage/storage-mq/src/storage-content/postgresStorageContentRepository.ts
```diff
@@ -43,28 +43,26 @@ export class PostgresStorageContentRepository implements StorageContentRepositor
   async existsTable (tableIdentifier: string): Promise<boolean> {
     const resultSet =
       await this.postgresRepository.executeQuery(EXISTS_TABLE_STATEMENT(POSTGRES_SCHEMA, tableIdentifier), [])
-    const tableExists = !!resultSet.rows[0].to_regclass
-    return Promise.resolve(tableExists)
+    return !!resultSet.rows[0].to_regclass
```
georg-schwarz: I would readd the explanatory variabale `tableExists`. Using these sort of unnecessary variables on the first glance with speaking name improves the readability of the code quite heavily in my opinion. Probably better named `foundTableWithIdentifier` or something similar since `tableExists` is duplicate with method name

sonallux: Ok, I have readded the variable.

sonallux:
I have rebased this PR onto the latest master to check if Mathias fix actually resolves the failing Adapter IT test.

adapter/integration-test/src/adapter-datasource-trigger.test.js
```diff
@@ -17,7 +17,7 @@ const EXECUTION_TOPIC = process.env.AMQP_IMPORT_TOPIC
 const EXECUTION_SUCCESS_TOPIC = process.env.AMQP_IMPORT_SUCCESS_TOPIC
 const EXECUTION_FAILED_TOPIC = process.env.AMQP_IMPORT_FAILED_TOPIC
 
-const STARTUP_DELAY = 2000
+const STARTUP_DELAY = 5000
```
mathiaszinnen, line 20: Why did you raise the delay? Did the adapter it still fail?

mathiaszinnen (approved):

adapter/integration-test/src/adapter-datasource-trigger.test.js
sonallux: Yes, adapter IT did still failed, see [this run](https://github.com/jvalue/open-data-service/runs/1086556642) of commit 2b2d219. I have also moved the sleep before connecting to amqp. According to the logs of the failing test the `Connected to AMQP on host ...` message from the [`connectAmqp()`](https://github.com/jvalue/open-data-service/blob/30a7be612be62bf85a99164df5c4ee380188699b/adapter/integration-test/src/adapter-datasource-trigger.test.js#L185) method was never outputed to the logs.

PR #194, opened 2020-09-15
Remove obsolete and useless code

sonallux:
- Remove obsolete usage of NodeJS `--max-http-header-size` parameter
- Remove useless wrapping of the return value with `Promise.reject(...)` or `Promise.resolve(...)` in `async` functions
- Move `sleep()` function to its own file, so it is not redefined multiple times

### DIFF
--- a/storage/storage-mq/src/sleep.ts
+++ b/storage/storage-mq/src/sleep.ts
@@ -1,0 +1,7 @@
+/**
+ * Returns a Promise that waits for a specific time period (in seconds) and resolves afterwards
+ * @param ms Period to wait in seconds
+ */
+export function sleep (ms: number): Promise<void> {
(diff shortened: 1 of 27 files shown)


storage/storage-mq/src/sleep.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Returns a Promise that waits for a specific time period (in seconds) and resolves afterwards
+ * @param ms Period to wait in seconds
```
georg-schwarz: in milliseconds!

georg-schwarz (changes requested):

storage/storage-mq/src/sleep.ts
sonallux: Oh, I have fixed it.

georg-schwarz (approved):
nice!

PR #207, opened 2020-09-29
Refactor adapter it

mathiaszinnen:


andreas-bauer (changes requested):
Please have a look at PR #203 and refactor it the same way.

andreas-bauer:
I did the following changes:
- throw error if in `waitForserviceToBeReady()` which shows the not reachable service on a timeout
- changed names of test (Should ...) in `adapter-stateless.test.js` to communicate the purpose of the tests better.
- console.log() of the ENVs as a single log (to avoid clutter through multiple console.log statements )

andreas-bauer (changes requested):

adapter/integration-test/src/adapter-datasource.test.js
```diff
@@ -160,19 +147,19 @@ describe('Adapter Configuration', () => {
     expect(eventsAfter.body[0].eventId).toBe(eventId + 1)
     expect(eventsAfter.body[0].datasourceId).toBe(datasourceId)
     expect(eventsAfter.body[0].eventType).toEqual('DATASOURCE_DELETE')
-  })
+  }, TIMEOUT)
 
   test('GET datasources/events/latest', async () => {
```
mathiaszinnen, line 152: I did not change the descriptions for the event tests since I am going to remove them in the next pull request anyways

andreas-bauer: Do you remove all tests?

mathiaszinnen: I want to replace the http event publishing by amqp. Therefore all tests related to the event http endpoints will be replaced by rabbitmq stuff

adapter/integration-test/src/adapter-datasource-trigger.test.js
```diff
@@ -151,22 +145,22 @@ describe('Adapter Sources Trigger', () => {
...
       .send(brokenDatasourceConfig)
 
     const datasourceId = datasourceResponse.body.id
 
-    const triggerResponse = await request(URL)
+    const triggerResponse = await request(ADAPTER_URL)
       .post(`/datasources/${datasourceId}/trigger`)
       .send()
 
     expect(triggerResponse.status).toBeGreaterThan(300) // request should fail (no 2xx status)
```
andreas-bauer: What status code is here really expected?
It can't be the case that a random error status code is returned.

mathiaszinnen: The status code returned is 404 which is not the behavior we want. IMO it should be 500. The root cause is addressed in issue #156. 
I agree that, when #156 is done, we should specify the response code. Until then, however, I would like to leave it open, since expecting 404 is not correct either.

adapter/integration-test/src/adapter-datasource-trigger.test.js
```diff
@@ -158,7 +157,7 @@ describe('Datasource triggering', () => {
       .post(`/datasources/${datasourceId}/trigger`)
       .send()
 
-    expect(triggerResponse.status).toBeGreaterThan(300) // request should fail (no 2xx status)
+    expect(triggerResponse.status).toBeGreaterThan(300)
```
mathiaszinnen: I don't think this comment is unnecessary. It is not immediately obvious that this assertion is meant to ensure that the request is supposed to fail, the comment makes it easier to understand this purpose

andreas-bauer: Than the test description or the expect() method should reflect that behavior and not a comment.
IMHO it is a code smell if a comment next to a statement has to explain this statement.
For example
```
expect(triggerResponse.status).toBeEqual(NOT_FOUND)
```

If it is in a `runInBand` configuration
```

test('Should respond with NOT_FOUND for not existing XYZ ', () => {
expect(triggerResponse.status).toBeEqual(404)
})
```

andreas-bauer (approved):

Issue #195, opened 2020-09-17
Proposal: Update linter rules


sonallux:
Currently the eslint configs from [eslint-config-standard](https://github.com/standard/eslint-config-standard) and [@typescript-eslint/recommanded](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin) are used. Using these two configs has a few drawbacks:

- The rules from `eslint-config-standard` are only designed for plain JavaScript, therefore there are some style inconsistencies with TypeScript (for example semicolons are prohibited everywhere except in interface definitions as they are only available in TypeScript.)
- `@typescript-eslint` rules that would require types checking are currently not enabled, because they are only available in the `@typescript-eslint/recommended-requiring-type-checking` config. Those rules would enforce stricter type checking and would spot many of the current present TypeScript issues in the codebase (Issues similar to #171).

### Possible solutions
1.  Add the `@typescript-eslint/recommended-requiring-type-checking` rule and manually adjust some `eslint` and `@typescript-eslint` rules to align with the `eslint-config-standard`.
    - Stricter rules for types than solution 2.: higher migration cost as some places would need a lot of refactoring
    - Manual configuration of the `@typescript-eslint` rules to align with `eslint-config-standard`

2. Use the [eslint-config-standard-with-typescript](https://github.com/standard/eslint-config-standard-with-typescript). This eslint config is based on the `eslint-config-standard` and enables TypeScript specific rules from `@typescript-eslint`.
    - Less strict rules for types than solution 1., but stricter than the current configuration: lower migration cost, because of less refactoring
    - No manual rule configuration and alignment with `eslint-config-standard`

I would prefer solution 2. as it would be easier to update and we do not need to maintain our config to align with `eslint-config-standard`. If stricter type checking rules from `@typescript-eslint` are desired, we can still add them manually to your configuration.

### Update plan
I think the update must be done in multiple steps, even if updating only one microservice at a time. In each step, I would manually adjust some rules that would align your eslint config more with the desired solution. For example, the first step would be to disable semicolons in interfaces. After a few iterations, we can update your config to only use the rules from the predefined configurations.

As this would be a rather big update, not for features but definitely for code quality, I would like to hear your opinions about my proposal. Feedback and/or suggestions are also very welcome. @georg-schwarz @mathiaszinnen @andreas-bauer @lunedis @jenswaechtler

georg-schwarz:
Agreed, that an iterative process is the right way to go here!

[...]

PR #211, opened 2020-10-02
Scheduler: Update linter config to standard-with-typescript

sonallux:
See #195


scheduler/src/clients/adapter-client.ts
```diff
@@ -27,7 +27,7 @@ export async function getEventsAfter (eventId: number): Promise<DatasourceEvent[
 
 export async function getLatestEventId (): Promise<number> {
   const response = await http.get('datasources/events/latest')
-  return response.data.eventId || 0
+  return response.data.eventId ?? 0
```
georg-schwarz, line 30: Very nice, didn't know that syntax.
I'll mention the others so they also see it!

@lunedis @andreas-bauer @mathiaszinnen

scheduler/src/index.ts
```diff
@@ -33,49 +33,38 @@ const server = app.listen(port, async () => {
...
     CHRONJOB_EVERY_2_SECONDS,
-    updateDatsources)
+    updateDatasources)
 }
 
-// log all promise rejections that happen
-// (mostly because they happen in async and don't log the point where it happened)
-// For more information see https://nodejs.org/api/process.html#process_event_unhandledrejection
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-process.on('unhandledRejection', (reason: Error | any, promise: Promise<any>) => {
```
georg-schwarz, line 67: Why remove it?
I would agree that it promotes not handling errors correctly which is not something we should encourage...

scheduler/src/clients/adapter-client.ts
andreas-bauer: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

scheduler/src/index.ts
sonallux: It should not be needed anymore with my changes. This was needed because in some places a function with a return type of `Promise<void>` was passed into a callback of type `(...) => void` which obviously does not handle the returned Promise:
https://github.com/jvalue/open-data-service/blob/5c24de8e6741d6bc5259f01c21df88571fc1110a/scheduler/src/scheduling.ts#L142-L144

The new linter rules will not allow this and will force everyone to handle the Promise appropiate. This means the rejected Promise case must be handled and the easiest way is to just add a `.catch(error => ...)` statement:
```ts
  const scheduledJob = schedule.scheduleJob(`Datasource ${datasourceId}`, executionDate, () => {
    execute(datasourceConfig)
      .catch(error => console.log('Failed to execute job:', error))
  })
```

georg-schwarz: Nice, thanks for the explanation!

georg-schwarz (approved):

PR #215, opened 2020-10-06
Notification: Remove ORM, refactor API & API model

georg-schwarz:
* [x] Remove ORM and replace with SQL statements
* [x] Change API model to follow type + parameter pattern
* [x] Change API paths (since type now in API model instead)
* [x] Update documentation
* [x] Change UI API consumption


notification/src/notification-config/notificationConfig.ts
```diff
@@ -1,57 +1,98 @@
...
-
-  @Column()
-  condition!: string
+export interface NotificationConfig {
+  id: number
+  pipelineId: number
+  condition: string
+  type: NotificationType
+  parameter: NotificationParameter
 }
```
sonallux: How about changing the model to this, because your current model does not force consistency between `type` and `parameter` (e.g. a NotificationConfig can have a slack notification type but webhook parameters)? With my suggestion, also some typecasts will be unnecessary and can be removed. 

```suggestion
export type NotificationConfig = SlackNotification | WebhookNotification | FirebaseNotification

export interface BaseNotification {
  id: number
  pipelineId: number
  condition: string
}

export interface SlackNotification extends BaseNotification{
  type: NotificationType.SLACK
  parameter: SlackParameter
}

export interface WebhookNotification extends BaseNotification{
  type: NotificationType.WEBHOOK
  parameter: WebhookParameter
}

export interface FirebaseNotification extends BaseNotification{
  type: NotificationType.FCM
  parameter: FirebaseParameter
}
```

georg-schwarz: Good point! I'm currently struggling to make it pretty :/ 
Another solution is here, which is much to complex IMHO: https://stackoverflow.com/questions/56949513/typescript-type-of-a-property-dependent-on-another-property-within-the-same-obj

There are a few switch-case statements in the code about the type and linked behavior. I'd like to get rid of that... probably by using objects? Ah it would be so nice to be able to overload methods with different subtypes and the compiler chooses the right one to execute...

sonallux: The other solutions from StackOverflow are actually the same base idea, just with fancier TypeScript syntax.

Well, that's the way of doing it in a functional programming language. But what is not pretty in your opinion? I have just played around a bit and came up with the following: https://github.com/jsone-studios/open-data-service/commit/8213019250b7aad341ebf8b138a753f179b2da85

georg-schwarz: Well, it bugs me that we have multiple places in the code with switch-case statements that you have to go over when adding e.g. another notification config. But nevermind, it's not that big of a deal.

I tried to simplify some other stuff here and there and even got rid of one switch-case in the database repository.

doc/example-requests/notification.http
```diff
@@ -4,38 +4,48 @@
...
+Content-Type: application/json
+
 ### Get all notification configs for pipeline
-GET {{baseURL}}/config/pipeline/17  HTTP/1.1
+GET {{baseURL}}/configs?pipelineId=17  HTTP/1.1
 Content-Type: application/json
 
 ### Save notification webhook config
-POST {{baseURL}}/config/webhook  HTTP/1.1
+POST {{baseURL}}/configs  HTTP/1.1
```
mathiaszinnen, line 16: example requests for the other types (slack, firebase) would be nice as well

notification/integration-test/src/notification-config.test.js
```diff
@@ -77,62 +74,93 @@ describe('Notification', () => {
...
+        secret: slackConfig.parameter.secret
+      }
     })
 
     // CLEANUP
     notificationResponse = await request(URL)
-      .delete(`/config/webhook/${id}`)
+      .delete(`/configs/${webhookId}`)
       .send()
     expect(notificationResponse.status).toEqual(200)
```
mathiaszinnen, line 165: You forgot deleting the slack webhook :)

notification/src/notification-execution/notificationExecutor.ts
```diff
@@ -29,61 +27,61 @@ export default class NotificationExecutor {
...
+    configParameter: WebhookParameter, dataLocation: string, message: string
+  ): Promise<void> {
+    const payload: WebhookCallback = {
       location: dataLocation,
       message: message,
       timestamp: new Date(Date.now())
     }
-    console.log(`Posting webhook to ${webhook.url}, callback object: ${JSON.stringify(callbackObject)}.`)
-    await axios.post(webhook.url, callbackObject)
+    await axios.post(configParameter.url, payload)
```
mathiaszinnen: we should probably add a catch clause to handle (log) failing requests here

notification/src/postgresRepository.ts
```diff
@@ -0,0 +1,70 @@
...
+      try {
+        return await this.connect(poolConfig)
+      } catch (error) {
+        lastError = error
+        console.info(`Error connecting to database (${i}/${retries})`)
+      }
+      await sleep(backoffMs)
+    }
+    // can be simplified in one line when https://github.com/typescript-eslint/typescript-eslint/issues/2642 is fixed
+    const error = lastError ?? new Error('Failed to connect to database')
```
mathiaszinnen: throw lastError ?? new Error('Failed to connect to database'). 
Jonas fixed it :D

notification/src/notification-execution/notificationExecutor.ts
```diff
@@ -29,61 +27,61 @@ export default class NotificationExecutor {
...
+  private async executeSlack (configParameter: SlackParameter, dataLocation: string, message: string): Promise<void> {
+    const payload: SlackCallback = {
       text: message
     }
-    const url = `${SLACK_BASE_URL}/${slack.workspaceId}/${slack.channelId}/${slack.secret}`
-    console.log(`Posting slack notification to ${url}, callbackObject: ${JSON.stringify(callbackObject)}`)
-    await axios.post(url, callbackObject)
+    const url =
+      `${SLACK_BASE_URL}/${configParameter.workspaceId}/${configParameter.channelId}/${configParameter.secret}`
+    await axios.post(url, payload)
```
mathiaszinnen: here too

notification/src/notification-execution/notificationExecutor.ts
```diff
@@ -93,10 +91,9 @@ export default class NotificationExecutor {
       data: {
         message: message
       },
-      topic: firebaseConfig.topic
+      topic: configParameter.topic
     }
-    console.log(`Sending firebase message, callback object: ${JSON.stringify(firebaseMessage)}.`)
-    const firebaseResponse = await firebase.messaging(app).send(firebaseMessage)
-    console.log(`Firebase message sent to: ${firebaseResponse}`)
+    await firebase.messaging(app).send(firebaseMessage)
```
mathiaszinnen: and here

mathiaszinnen (dismissed):
LGTM

notification/src/postgresRepository.ts
georg-schwarz: Hm, kinda still get the error :(

image

sonallux: It should work with the latest version of typescript-eslint: `4.4.1`. VSCode may still use the old version because it did not update its internal linter state correctly after installing the new typescript-eslint version. When manually running `npm run lint` the error should not be visible.

mathiaszinnen (dismissed):

notification/src/postgresRepository.ts
georg-schwarz: Thanks, `npm i` solved it..

notification/src/postgresRepository.ts
```diff
@@ -31,8 +31,7 @@ export default class PostgresRepository {
       await sleep(backoffMs)
     }
     // can be simplified in one line when https://github.com/typescript-eslint/typescript-eslint/issues/2642 is fixed
```
sonallux:
```suggestion
```
The comment can also be removed

mathiaszinnen (approved):

PR #217, opened 2020-10-07
Storage-mq: Update linter config to standard-with-typescript

sonallux:
See #195


storage/storage-mq/src/api/pipelineExecutionEventHandler.ts
```diff
@@ -8,17 +8,17 @@ export default class PipelineExecutionEventHandler {
   }
 
   async handleSuccess (pipelineExecutedEvent: PipelineExecutedEvent): Promise<void> {
-    await this.contentRepository.saveContent(pipelineExecutedEvent.pipelineId, {
+    await this.contentRepository.saveContent(pipelineExecutedEvent.pipelineId.toString(), {
       pipelineId: pipelineExecutedEvent.pipelineId,
-      timestamp: pipelineExecutedEvent.timestamp,
-      data: pipelineExecutedEvent.data,
-      id: undefined
+      timestamp: pipelineExecutedEvent.timestamp ?? new Date(),
```
sonallux, line 13: This is kind of a workaround because the messages from the integration tests contain the `timestamp` field, but the actual events from the Pipeline service do not: https://github.com/jvalue/open-data-service/blob/master/pipeline/src/pipeline-config/publisher/amqpExecutionResultPublisher.ts#L30.

This is just one instance of a broader issue/difficulty that exists in the codebase: The type definitions of the amqp messages are always duplicated at producer and consumer site. Which will lead to diverging type definitions if not all producers and consumers are updated simultaneously.

storage/storage-mq/src/storage-content/postgresStorageContentRepository.ts
```diff
@@ -41,10 +53,10 @@ export class PostgresStorageContentRepository implements StorageContentRepositor
   }
 
   async existsTable (tableIdentifier: string): Promise<boolean> {
-    const resultSet =
+    const resultSet: QueryResult<ExistsTableResultRow> =
       await this.postgresRepository.executeQuery(EXISTS_TABLE_STATEMENT(POSTGRES_SCHEMA, tableIdentifier), [])
-    const foundTableWithIdentifier = !!resultSet.rows[0].to_regclass
-    return foundTableWithIdentifier
+    // to_regclass contains the table name as string or null if it does not exists
```
georg-schwarz: Was already present in the codebase, but probably would be cleaner to extract to method with a speaking name... would you mind including that in your PR?

storage/storage-mq/src/api/pipelineExecutionEventHandler.ts
georg-schwarz: I totally agree - that's one of the challenges with a distributed architecture. We won't want to have shared libs for that because microservices want to avoid that (we are not there to enjoy the benefits of that)... When we introduce consumer-driven contract tests this should be solved. Until now, we have our system tests to make sure that the main functionality works... Details like that are not tested, so we have to live with these little mismatches for now ...

storage/storage-mq/src/storage-content/postgresStorageContentRepository.ts
sonallux: Already done

georg-schwarz (approved):
Thanks!

PR #219, opened 2020-10-08
Adapter scheduler amqp

mathiaszinnen:


adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/api/amqp/AmqpHandler.java
```diff
@@ -0,0 +1,63 @@
...
+import org.jvalue.ods.adapterservice.datasource.model.Datasource;
+import org.springframework.amqp.AmqpException;
+import org.springframework.amqp.rabbit.core.RabbitTemplate;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+import java.io.Serializable;
+
+@Service
+public class AmqpHandler {
```
georg-schwarz: Rename to AmqpPublisher?
Divide into two classes, for config and import? (not necessary)

adapter/integration-test/src/adapter-datasource.test.js
```diff
@@ -1,49 +1,75 @@
...
       .send(datasourceConfig)
+    const datasourceId = response.body.id
 
     expect(response.status).toEqual(201)
     expect(response.header.location).toContain(response.body.id)
     expect(response.body.adapter).toEqual(datasourceConfig.adapter)
     expect(response.body.trigger).toEqual(datasourceConfig.trigger)
     expect(response.body.id).toBeDefined()
-    expect(response.body.id).not.toEqual(datasourceConfig.id) // id not under control of client
+    expect(datasourceId).not.toEqual(datasourceConfig.id) // id not under control of client
```
andreas-bauer: If the `id` is given by the server `datasourceConfig` should not have a reasonable`id`.
Possible values could be `undefined`, `0`, or `-1`.
So you could test if a positive number was given as `id`.

In the current case a `-1234` value would be a pass, but should be a fail.

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/api/amqp/AmqpPublisher.java
```diff
@@ -0,0 +1,61 @@
...
+                this.rabbitTemplate.convertAndSend(RabbitConfiguration.AMPQ_EXCHANGE, topic, message);
+                return;
+            } catch (AmqpException e) {
+                try {
+                    Thread.sleep(RabbitConfiguration.AMQP_PUBLISH_BACKOFF);
+                } catch (InterruptedException interruptedException) {
+                    Thread.currentThread().interrupt();
+                    throw new RuntimeException(interruptedException);
+                }
+                System.out.println("Message publish failed ("+retries+"). Retrying in "+RabbitConfiguration.AMQP_PUBLISH_BACKOFF);
```
georg-schwarz: Only log in debug mode?

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/event/DatasourceCreationEvent.java
```diff
@@ -0,0 +1,41 @@
+package org.jvalue.ods.adapterservice.datasource.event;
+
+import org.jvalue.ods.adapterservice.datasource.model.Datasource;
+import org.jvalue.ods.adapterservice.datasource.model.DatasourceTrigger;
+
+import java.util.Objects;
+
+public class DatasourceCreationEvent extends DatasourceConfigEvent {
+
+    private final DatasourceTrigger trigger;
```
georg-schwarz: Probably it is better to send the whole config (?)
Clients will only extract the parts they need then (e.g. the scheduler extracts the trigger config)

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/event/DatasourceUpdateEvent.java
```diff
@@ -0,0 +1,41 @@
+package org.jvalue.ods.adapterservice.datasource.event;
+
+import org.jvalue.ods.adapterservice.datasource.model.Datasource;
+import org.jvalue.ods.adapterservice.datasource.model.DatasourceTrigger;
+
+import java.util.Objects;
+
+public class DatasourceUpdateEvent extends DatasourceConfigEvent {
```
georg-schwarz: Exact duplicate of CreateEvent? => then combine to one

adapter/integration-test/src/adapter-datasource.test.js
georg-schwarz: You mean we should return `400 Bad Request: Field id should not be set` instead of just ignoring it?

mathiaszinnen: I will change the adapter service so it does not accept post requests with specified id fields

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/api/amqp/AmqpPublisher.java
mathiaszinnen: You mean using system.err or using the spring logging framework?

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/event/DatasourceCreationEvent.java
mathiaszinnen: Good point, will change it accordingly

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/event/DatasourceUpdateEvent.java
mathiaszinnen: Right, this is redundant. I will simplify the config events by using only one for creation, update and deletion.

scheduler/src/api/amqp/datasourceConfigConsumer.ts
```diff
@@ -0,0 +1,73 @@
...
+    await channel.bindQueue(q.queue, AMQP_SCHEDULER_EXCHANGE, AMQP_DATASOURCE_CONFIG_TOPIC)
+
+    await channel.consume(q.queue, msg => {
+      this.consumeEvent(msg)
+        .catch(error => console.error(`Failed to handle ${msg?.fields.routingKey ?? 'null'} event`, error))
+    })
+    console.info('Successfully initialized AMQP queue')
+  }
+
+  consumeEvent = async (msg: AMQP.ConsumeMessage | null): Promise<void> => {
```
andreas-bauer, line 60: You should try to avoid nested control flow structures.
Also the if conditions are to complex and should be extracted into separate methods, like `isDeleteTopic(msg)`.

```js
if (msg === null) {
  console.debug(...)
  return // or return Promise.reject(reason)
}

console.debug("[ConsumingEvent]...")

if (isUpdateTopic(msg)) {
  ...
}

if (...) {
  ...
}
```

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/api/amqp/AmqpPublisher.java
georg-schwarz: No, just thinking in general how to handle that error... Printing seems to be a good way since we do retries. There, we have a few log levels in general (debug, info, warning, error) where we can indicate stuff. Probably we want do define when to use which log levels in the future. Nvm for now.

scheduler/src/api/amqp/datasourceConfigConsumer.ts
```diff
@@ -0,0 +1,85 @@
...
+    if (msg === null) {
+      console.debug('Received empty event when listening on datasource config events - doing nothing')
+    } else {
+      await this.handleMsg(msg)
+    }
+  }
+
+  handleMsg = async (msg: AMQP.ConsumeMessage): Promise<void> => {
+    console.debug("[ConsumingEvent] %s:'%s'", msg.fields.routingKey, msg.content.toString())
+    if (isUpdateOrCreate(msg)) {
```
andreas-bauer, line 71: The if statement is still not easy to understand when browsing the code.

A refactoring like this one would improve the readability of the code.
```js
if (isUpdateOrCreate(msg)) {
    await this.scheduler.applyCreateOrUpdateEvent(JSON.parse(msg.content.toString()))
    return
}

if (isDelete(msg)) {
    this.scheduler.applyDeleteEvent(JSON.parse(msg.content.toString()))
    return
}

console.debug('...')
```

scheduler/src/scheduling.test.ts
```diff
@@ -1,111 +1,105 @@
...
+    await sleep(500)
 
-    await sleep(1000) // Give the scheduler some time to apply the update
-
-    expect(Scheduling.getAllJobs()).toHaveLength(1)
-    const job123 = Scheduling.getJob(123)
+    expect(scheduler.getAllJobs()).toHaveLength(1)
+    const job123 = scheduler.getJob(123)
     expect(job123).toBeDefined()
     if (job123 !== undefined) {
```
andreas-bauer: A if statement which exclude a `expect()` statement should not be present in a test.

`expect(job123).toBeDefined()` checks already if the job is not `undefined`. 

I would remove the if statement and just write the check inside the `{}` direct after the other.

scheduler/src/scheduling.test.ts
```diff
@@ -1,111 +1,105 @@
...
 
-    await sleep(1000) // Give the scheduler some time to apply the update
+    await scheduler.applyCreateOrUpdateEvent(updateEvent)
 
-    expect(Scheduling.getAllJobs()).toHaveLength(1)
-    const job123 = Scheduling.getJob(123)
+    expect(scheduler.getAllJobs()).toHaveLength(1)
+    const job123 = scheduler.getJob(123)
     expect(job123).toBeDefined()
     if (job123 !== undefined) {
```
andreas-bauer: See comment above.

scheduler/src/scheduling.ts
```diff
@@ -3,217 +3,179 @@ import deepEqual from 'deep-equal'
...
-  })
-  allJobs.clear()
+const isAxiosError = function (error: any): error is AxiosError {
+  return error.isAxiosError
 }
 
-export function cancelJob (jobId: number): void {
-  const job = allJobs.get(jobId)
-  job?.scheduleJob.cancel()
+const handleAxiosError = function (error: AxiosError): void {
```
andreas-bauer, line 168: Also this method lacks of easy readability because of the if statements.

Here is a refactoring proposal.

```js
 const handleAxiosError = function (error: AxiosError): void {
    const baseMsg = 'Error during datasource triggering: ' 
    if (error.response !== undefined) {
      console.error(`${baseMsg} [${error.response.status}] ${error.response.data}`)
      return
    } 
    
    if (error.request !== undefined) {
        console.error(`${baseMsg} Request failed: ${JSON.stringify(error.request)}`)      
        return
    }

    console.error(`${baseMsg} unknown reason`)    
  }
```

scheduler/src/scheduling.test.ts
mathiaszinnen: I think the if-clause was added because the typescript compiler does not recognize the non-null check by `expect()`. I will suppress the ts error by `@ts-suppress.`

scheduler/src/scheduling.ts
mathiaszinnen: Why do you consider multiple `if ... return` statements more readable than `if ... else`? I don't have a strong opinion on that so I don't mind changing it, but I am curious anyway.

andreas-bauer: IMHO the `if ... return` (also known as Guard Clauses) make the code more lean and readable.
1) The blank line between the if statements helps to identify the different outcomes more easily. Whereas the if-else statement looks like a big blob. 
2) Readability does not decrease in the way it does with if-else statements when you add more checks.
3) You avoid unnecessary branching in complex scenarios.

mathiaszinnen: Alright, thanks for the clarification!

adapter/integration-test/src/adapter-datasource.test.js
```diff
@@ -1,77 +1,125 @@
...
+    }
+  }, TIMEOUT)
+
   test('Should respond with stored datasource configs', async () => {
     const response = await request(ADAPTER_URL).get('/datasources')
     expect(response.status).toEqual(200)
     expect(response.type).toEqual('application/json')
 
-    expect(response.body).toEqual([])
+    expect(response.body).toBeInstanceOf(Array)
```
georg-schwarz, line 47: Well, that test is non-saying now.
We should either test that a created one is contained, or that it is empty if no data sources are present. I'd prefer the first option!

adapter/integration-test/src/adapter-datasource.test.js
```diff
@@ -1,77 +1,125 @@
...
   }, TIMEOUT)
 
+  test('Should reject datasources with specified id [POST /datasources]', async () => {
```
georg-schwarz: remove the URL because it can change. Rather say "when creating a datasource"

adapter/integration-test/src/adapter-datasource.test.js
```diff
@@ -1,77 +1,125 @@
...
+    })
   }, TIMEOUT)
 
   test('Should update existing datasource [PUT /datasources/{id}]', async () => {
     const postResponse = await request(ADAPTER_URL)
       .post('/datasources')
       .send(datasourceConfig)
-
-    const datasourceId = postResponse.body.id
+    const datasource = postResponse.body
```
georg-schwarz: better name createdDatasource. It's hard to tell what the difference between datasource and datasourceConfig.

adapter/integration-test/src/adapter-datasource.test.js
```diff
@@ -1,77 +1,125 @@
...
+      .post('/datasources')
+      .send(datasourceConfig)
+    const datasource = response.body
 
     const delResponse = await request(ADAPTER_URL)
-      .delete('/datasources/' + datasourceId)
+      .delete('/datasources/' + datasource.id)
       .send()
 
     expect(delResponse.status).toEqual(204)
```
georg-schwarz, line 122: Add check if GET returns 404 after deletion?

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/DatasourceManager.java
```diff
@@ -64,30 +53,31 @@ public Datasource createDatasource(Datasource config) {
...
 
   @Transactional
   public void deleteAllDatasources() {
     Iterable<Datasource> allDatasourceConfigs = getAllDatasources();
     datasourceRepository.deleteAll();
-
-    allDatasourceConfigs.forEach(
-      pl -> eventRepository.save(new DatasourceEvent(EventType.DATASOURCE_DELETE, pl.getId()))
-    );
+    for (Datasource ds: allDatasourceConfigs) {
```
georg-schwarz: use a stream instead, otherwise one unsuccessful deletion publish might block the other ones

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/api/amqp/AmqpPublisher.java
```diff
@@ -0,0 +1,61 @@
...
+                this.rabbitTemplate.convertAndSend(RabbitConfiguration.AMPQ_EXCHANGE, topic, message);
+                return;
+            } catch (AmqpException e) {
+                try {
+                    Thread.sleep(RabbitConfiguration.AMQP_PUBLISH_BACKOFF);
+                } catch (InterruptedException interruptedException) {
+                    Thread.currentThread().interrupt();
+                    throw new RuntimeException(interruptedException);
+                }
+                System.err.println("Message publish failed ("+retries+"). Retrying in "+RabbitConfiguration.AMQP_PUBLISH_BACKOFF);
```
georg-schwarz: Better log level warning. Error will occur if retries don't help.

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/api/rest/v1/DatasourceEndpoint.java
```diff
@@ -39,7 +39,9 @@ public Datasource getDatasource(@PathVariable Long id) {
 
     @PostMapping
     public ResponseEntity<Datasource> addDatasource(@Valid @RequestBody Datasource config) {
-          config.setId(null); // id not under control of client
+        if (config.getId() != null && config.getId() > 0) {
```
georg-schwarz: so negative ids allowed? I'd rather just not allow any id.

scheduler/README.md
```diff
@@ -4,19 +4,16 @@
...
-The server should now be running on http://localhost:8080
+Since the scheduler performs an initial sync with the adapter service, it is currently only available to start in in the docker environment:
+Run `docker-compose -f ../docker-compose.yml up scheduler adapter`. 
 
 ### Testing
 
 Run the unit tests with `npm test`. Jest is used as unit testing framework.
 
 For integration testing run `docker-compose -f ../docker-compose.yml -f ../docker-compose.ci.yml up scheduler-service scheduler-service-it`.
+To run integration tests outside of the docker environment run the `it_local.sh` script.
```
georg-schwarz, line 15: I think we should discuss about this script. Is it something we want everywhere? Do we better provide env files instead?

scheduler/integration-test/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "mock:adapter": "node src/mock.adapter.js",
     "pretest": "npm run mock:adapter &",
-    "test": "jest",
+    "test": "jest --runInBand",
```
georg-schwarz, line 8: Necessary?

scheduler/integration-test/src/scheduler.test.js
```diff
@@ -1,52 +1,176 @@
...
+  throw new Error('Could not establish connection to AMQP broker')
+}
+
+const createAmqpChannel = async () => {
+  const channel = await amqpConnection.createChannel()
+  await channel.assertExchange(AMQP_EXCHANGE, 'topic', { durable: true })
+  return channel
+}
+
+const logConfigs = () => {
```
georg-schwarz, line 162: Should be done in env file

scheduler/src/api/amqp/datasourceConfigConsumer.ts
```diff
@@ -0,0 +1,90 @@
...
+  consumeEvent = async (msg: AMQP.ConsumeMessage | null): Promise<void> => {
+    if (msg === null) {
+      console.debug('Received empty event when listening on datasource config events - doing nothing')
+    } else {
+      await this.handleMsg(msg)
+    }
+  }
+
+  handleMsg = async (msg: AMQP.ConsumeMessage): Promise<void> => {
+    console.debug("[ConsumingEvent] %s:'%s'", msg.fields.routingKey, msg.content.toString())
```
georg-schwarz: dependend on merge order: adapt tp #229

scheduler/src/interfaces/datasource-config-event.ts
```diff
@@ -0,0 +1,5 @@
+import DatasourceConfig from '@/interfaces/datasource-config'
```
georg-schwarz: That import should not work?

scheduler/src/interfaces/datasource-config.ts
```diff
@@ -1,16 +1,6 @@
...
-}
-
-export interface DatasourceProtocol {
-  type: string
-  parameters: {
-    location?: string
-  }
 }
 
 export interface DatasourceTrigger {
```
georg-schwarz, line 6: We are having a few files with only a model class. Maybe we can put them to the files where they are used / logically belong to? Only makes sense if these files are rather small ..

scheduler/src/scheduling.ts
```diff
@@ -3,217 +3,183 @@ import deepEqual from 'deep-equal'
...
-
-  const datasources: DatasourceConfig[] = await AdapterClient.getAllDatasources()
-
-  console.log(`Received ${datasources.length} datasources from adapter-service`)
-
-  for (const datasource of datasources) {
-    datasource.trigger.firstExecution = new Date(datasource.trigger.firstExecution)
-    await Scheduling.upsertJob(datasource) // assuming adapter service checks for duplicates
+export default class Scheduler {
+  constructor (triggerRetries: number) {
```
georg-schwarz: `constructor (private triggerRetries: number) { }`

scheduler/src/scheduling.ts
```diff
@@ -3,217 +3,183 @@ import deepEqual from 'deep-equal'
...
+  private readonly allJobs: Map<number, ExecutionJob> = new Map() // datasourceId -> job
+
+  async initializeJobsWithRetry (retries: number, backoff: number): Promise<void> {
+    for (let i = 1; i <= retries; i++) {
+      try {
+        await this.initializeJobs()
+        return
+      } catch (e) {
+        if (e.code === 'ECONNREFUSED' || e.code === 'ENOTFOUND') {
+          console.error(`Failed to sync with Adapter Service on init (${retries}) . Retrying after ${backoff}ms... `)
```
georg-schwarz: better a warning

scheduler/src/scheduling.ts
```diff
@@ -3,217 +3,183 @@ import deepEqual from 'deep-equal'
...
+        } else {
+          console.error(e)
+          console.error(`Retrying (${retries})...`)
```
georg-schwarz: warning

adapter/integration-test/src/adapter-datasource.test.js
mathiaszinnen: I changed this because I did not want this test to make assumptions about the initial adapter state. Specifically, I wanted to enable running the test locally without having to delete docker volumes first. 
However, maybe, having a stronger test is more important. What do you think?

adapter/integration-test/src/adapter-datasource.test.js
mathiaszinnen: That holds for the other tests as well, right? Do you think we should change all test names containing the endpoint?

scheduler/README.md
mathiaszinnen: Yeah, I think so too. I added this script for local testing, and I included it into the pr just in case somebody else may need it, too. I will create a discussion issue so we can find a common approach for all services. Maybe we can discuss it in the next dev meeting as well.

scheduler/integration-test/package.json
mathiaszinnen: Since the tests are heavily timing dependant (which is problematic anyways), I found that the `runInBand` flag imporoved the stability a little. So I would keep it for now, yes. Of course, it would be better if we found a more robust way to test the scheduler but I did not have an idea yet.

scheduler/integration-test/src/scheduler.test.js
mathiaszinnen: Since the AMQP topics and the exchange name are defined in the test, this is not possible. 
Of course, we could also define them in the env vars, but I think this should be done in another pr. Especially since that would affect the other services' integration tests as well.

andreas-bauer: I would also keep it in the `scheduler.test.js` file. This way it will only produce one console log output for the test. In previous solutions where the ENVs are logged from the `env.js` one by one it produced a lot of clutter.

scheduler/src/scheduling.test.ts
```diff
@@ -1,121 +1,113 @@
...
 
-    expect(Scheduling.getAllJobs()).toHaveLength(1)
-    const job123 = Scheduling.getJob(123)
+    expect(scheduler.getAllJobs()).toHaveLength(1)
+    const job123 = scheduler.getJob(123)
     expect(job123).toBeDefined()
-    if (job123 !== undefined) {
-      expect(job123.datasourceConfig).toEqual(toBeAdded)
-    }
+    // @ts-expect-error
```
andreas-bauer: I looked again at this problem. It seams like Optional Chaining is the better solution instead of the suppression of the TypeScript warning.

```js
    expect(job123?.datasourceConfig).toEqual(toBeAdded)
    // if job123 or the attribute datasourceConfig is not defined it will return 'undefined'
```

scheduler/src/scheduling.test.ts
```diff
@@ -1,121 +1,113 @@
...
 
-    expect(Scheduling.getAllJobs()).toHaveLength(1)
-    const job123 = Scheduling.getJob(123)
+    expect(scheduler.getAllJobs()).toHaveLength(1)
+    const job123 = scheduler.getJob(123)
     expect(job123).toBeDefined()
-    if (job123 !== undefined) {
-      expect(job123.datasourceConfig).toEqual(updated)
-    }
+    // @ts-expect-error
```
andreas-bauer: I looked again at this problem. It seams like Optional Chaining is the better solution instead of the suppression of the TypeScript warning.

```js
    expect(job123?.datasourceConfig).toEqual(toBeAdded)
    // if job123 or the attribute datasourceConfig is not defined it will return 'undefined'
```

scheduler/src/scheduling.test.ts
```diff
@@ -138,102 +130,85 @@ describe('Scheduler', () => {
...
     expect(nextInvocation1).not.toEqual(nextInvocation2)
     expect(datasourceJob2).toBeDefined()
-    if (datasourceJob2 !== undefined) {
-      expect(datasourceJob1.datasourceConfig).toEqual(datasourceJob2.datasourceConfig)
-    }
-  })
-
-  afterEach(() => {
-    Scheduling.cancelAllJobs()
+    // @ts-expect-error
```
andreas-bauer: Please use Optional Chaining operator `myObj?.myAttribute`.

scheduler/integration-test/package.json
andreas-bauer: I also prefer the `--runInBand` option. Several people reported that `--runInBand` helps in an environment with limited resources, like a CI environment.

See [GitHub issue](https://github.com/facebook/jest/issues/1524#issuecomment-260246008)

adapter/integration-test/src/adapter-datasource.test.js
georg-schwarz: Can we add it to the create test? Like we create one, then use the get and check if it is contained in all configs?

mathiaszinnen: Yes, good idea, I will add that

georg-schwarz (approved):

PR #224, opened 2020-10-19
UI: limit transformation preview data to max height and scroll overflow

georg-schwarz:
Previously you had to scroll to the bottom of all data to further configure. Took a while with longer datasets. Limited to a max height now, and people can scroll through if they are interested.


lunedis:
Good idea! LGTM

lunedis (approved):

PR #226, opened 2020-10-20
WIP: Pipeline executer

jenswaechtler:
@georg-schwarz


pipeline-trigger/src/api/amqp/amqpHandler.ts
```diff
@@ -0,0 +1,81 @@
+// import * as AMQP from 'amqplib'
```
andreas-bauer: Why the content of this file is completely commented out?

pipeline-trigger/integration-test/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM node:lts-alpine as builder
+
+COPY ./src ./src
+COPY ./package*.json ./
+COPY ./.eslint* ./
+
+RUN npm ci
+RUN npm run lint-ci
```
andreas-bauer: To make use of the Docker image layer cache you need to structure the command in another way.

```
COPY ./package*.json ./
RUN npm ci

COPY ./src ./src
COPY ./.eslint* ./
RUN npm run lint-ci
```

pipeline-trigger/src/index.ts
```diff
@@ -0,0 +1,39 @@
...
+// import cors from 'cors'
+import bodyParser from 'body-parser'
+// import { AmqpHandler } from '@/api/amqp/amqpHandler'
+import { PipelineExecutor } from './pipeline-executor/pipeline-executor'
+import { DataImportEndpoint } from './api/rest/dataImportEndpoint'
+
+const port = 8080
+
+async function main (): Promise<void> {
+  // const amqpHandler = new AmqpHandler(triggerEventHandler)
```
andreas-bauer: Commented out code should not be present in the code.

pipeline-trigger/src/pipeline-executor/pipeline-executor.test.ts
```diff
@@ -0,0 +1,34 @@
...
+  ADAPTER_API: '_',
+  PIPELINE_API: '_'
+}))
+
+describe('PipelineExecutor', () => {
+  const getPipeline = PipelineClient.getPipeline as jest.Mock
+  const triggerDatasource = DatasourceClient.triggerDatasource as jest.Mock
+  const pipelineExecutor = new PipelineExecutor()
+
+  it('should execyte dataImportRequests correctly', async () => {
```
andreas-bauer: typo

pipeline-trigger/start_local.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+export ADAPTER_API=http://localhost:9000/api/adapter
+export PIPELINE_API=http://localhost:9000/api/pipelines
+
+npm run transpile
+node dist/index.js
```
andreas-bauer: This is usually done by by `.env` files. See ui folder.

pipeline-trigger/src/pipeline-executor/pipeline-executor.ts
```diff
@@ -0,0 +1,19 @@
...
+  get version (): string {
+    return this._version
+  }
+
+  async execute (dataImportRequest: DataImportRequest): Promise<void> {
+    const pipeline = await PipelineClient.getPipeline(dataImportRequest.pipelineId)
+
+    await DatasourceClient.triggerDatasource(pipeline.datasourceId)
+    console.log('triggered!')
+    console.log(pipeline.datasourceId)
```
andreas-bauer: These `console.log()` statements are not useful.
If you really want to communicate that a pipeline was triggered than the message should communicate this without room for interpretation. 

```js
console.log('Triggered pipeline for datasource with id ' + pipeline.datasourceId)
```

pipeline-trigger/.dockerignore
```diff
@@ -0,0 +1,92 @@
+# Created by https://www.gitignore.io/api/node
```
andreas-bauer: Why is the `.dockerignore` file necessary?
In the `Dockerfile`  you only copy the source and config files you need.

andreas-bauer:
@jenswaechtler There is a conflict with the following files. 
``` 
notification/package-lock.json
notification/package.json 
```

pipeline-trigger/.dockerignore
sonallux: @andreas-bauer This is absolutely necessary for better docker build performance. Without a `.dockerignore` the **whole** directory is sent to the Docker daemon regardless of the `COPY` commands in the `Dockerfile`. Therefore it is good practice to exclude for example the `node_modules` or `dist` folder from the docker build context (see https://docs.docker.com/engine/reference/builder/#usage for more details).

pipeline-trigger/start_local.sh
sonallux: For the microservices `.env` files are currently not supported, because using `.env` files would require additional tooling, which is provided by the VUE CLI out of the box.

Either the `.env` file must be loaded by the microservice itself or use the [dotenv](https://www.npmjs.com/package/dotenv) package and change the start command in `package.json` to `node -r dotenv/config --enable-source-maps dist/index.js`. 

But I would not make this change in this PR instead, I would do it in another PR for all microservices simultaneously to have a consistent developer experience across all microservices.

andreas-bauer: Thanks for the clarification. Looked like a hack at the first glance.

pipeline-trigger/src/api/amqp/amqpHandler.ts
jenswaechtler: I forgot to set the Pull Request as draft. 
I just changed it. 
Georg wanted to talk about the changes, so I made a pull request. 
It's not cleaned up yet. 
Thanks for the feedback, I will change it.

PR #227, opened 2020-10-21
Fix jest is ignoring paths option from tsconfig.json

sonallux:
- Currently, jest ignores the `paths` option from `tsconfig.json`, which causes test failures if a file needed for the test contains a `@/...` import. This PR adds a `moduleNameMapper` to `jest.config.js` to respect typescript's `paths` option.
- This PR changes absolute imports from `@/...` to `@ods/...` to be more aligned with the JS/TS community trend towards scoped packages (e.g. `@angular/...`, `@babel/...`, `@typescript-eslint/...`).
- Refactor and remove unused/useless options from `tsconfig.json`


sonallux:
TypeScript does not transform absolute imports into relative imports. Because NodeJS only allows relative imports, adding support for `@ods/...` imports do not work without major changes in tooling. Therefore I am closing this PR and creating a new one which removes the absolute imports with `@/...`.

PR #228, opened 2020-10-22
Remove absolute imports with `@/...`

sonallux:
Currently, our TypeScript config allows absolute imports with `@/...`. Those imports won't be transformed into relative imports by tsc. Because NodeJS does not understand these absolute imports with `@/...`, support for them is removed in this PR.


mathiaszinnen (commented):
Nice, this bugged me all the time

mathiaszinnen (approved):

georg-schwarz (approved):

PR #231, opened 2020-10-26
Adapter datasource validation

sonallux:
This is the first PR for #156. It only adds validation for the datasource protocol type and format type fields. This is achieved by adding a `Protocol` and a `Format` enum. The PR also removes the `AdapterFactory` to simplify the architecture.

Validation for the other parameters will be added in follow-up PRs.


sonallux:
@georg-schwarz @andreas-bauer @mathiaszinnen This is ready for review

georg-schwarz (dismissed):
LGTM - I'll let someone else look over it as well..

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/model/Datasource.java
```diff
@@ -11,123 +12,121 @@
...
-        triggerParameters = runtimeParameters.parameters;
-      }
-      Map<String, String> replacementParameters = Stream.of(defaultParameters, triggerParameters).flatMap(map -> map.entrySet().stream()).collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (v1, v2) -> v2));
-      for (Map.Entry<String, String> parameter : replacementParameters.entrySet()) {
-        url = url.replace("{" + parameter.getKey() + "}", parameter.getValue());
-      }
-      HashMap<String, Object> parameters = new HashMap<>(this.getProtocol().getParameters());
-      parameters.put("location", url);
-      return new DatasourceProtocol(this.getProtocol().getType(), parameters);
+  protected Map<String, Object> fillQueryParameters(RuntimeParameters runtimeParameters) {
```
andreas-bauer, line 95: This method looks very crowded.
Could you please add a blank line between the different `Map<>...` statements.

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/model/Datasource.java
```diff
@@ -11,123 +12,121 @@
...
+    String url = (String) this.getProtocol().getParameters().get("location");
+    Map<String, String> defaultParameters = new HashMap<>();
+    if (this.getProtocol().getParameters().containsKey("defaultParameters")) {
+      defaultParameters = (Map<String, String>) this.getProtocol().getParameters().get("defaultParameters");
     }
+    Map<String, String> triggerParameters = new HashMap<>();
+    if (runtimeParameters != null && runtimeParameters.parameters != null) {
+      triggerParameters = runtimeParameters.parameters;
+    }
+    Map<String, String> replacementParameters = Stream.of(defaultParameters, triggerParameters).flatMap(map -> map.entrySet().stream()).collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (v1, v2) -> v2));
```
andreas-bauer: This line is too long. You should always split stream API operation into separate lines.
```
Map<String, String> replacementParameters = Stream.of(defaultParameters, triggerParameters)
   .flatMap(map -> map.entrySet().stream())
   .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (v1, v2) -> v2));
```

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/Protocol.java
```diff
@@ -0,0 +1,18 @@
+package org.jvalue.ods.adapterservice.adapter;
+
+import org.jvalue.ods.adapterservice.adapter.importer.HttpImporter;
+import org.jvalue.ods.adapterservice.adapter.importer.Importer;
+import org.springframework.web.client.RestTemplate;
+
+public enum Protocol {
```
andreas-bauer, line 7: Will it work if someone uses lower case `http` or `Http` in the requesting REST API call?

andreas-bauer:
Can `Protocol` and `Format` be in lower case `http` or `json` in a REST request?

sonallux:
> Can `Protocol` and `Format` be in lower case `http` or `json` in a REST request?

No, but it has not worked with lower case before, so nothing has changed from the endpoint user perspective.

If we want to support lower case `http` or even `Http` then we can do that by using a custom Jackson deserializer. But that should be done in a separate PR.

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/model/Datasource.java
sonallux: Well, this method was there before, I have just adjusted the return type and changed the indentation from 4 to 2 spaces. But I will try to make this method more beautiful.

andreas-bauer: Hi, I know you didn't created the method. But when you touch it you can also leave it cleaner than before.
The Boy Scout Rule 🙃

andreas-bauer:
Ok, I was just curious. A custom Jackson deserializer just introducing more complexity. As long nobody complains we can keep it this way.

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/model/Datasource.java
sonallux: I have already refactored it and it is much cleaner now

andreas-bauer: Looks better now. Thanks

andreas-bauer (approved):

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/Protocol.java
mathiaszinnen: > > Can `Protocol` and `Format` be in lower case `http` or `json` in a REST request?
> 
> No, but it has not worked with lower case before, so nothing has changed from the endpoint user perspective.
> 
> If we want to support lower case `http` or even `Http` then we can do that by using a custom Jackson deserializer. But that should be done in a separate PR.

Couldn't we just extend the existing enum with the other spellings? I agree, however, that we do not need this feature and should leave it out anyway

mathiaszinnen (approved):
LGTM, too. Much cleaner than before

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/Protocol.java
sonallux: Yes, technically that works, too. But I would consider having enums with different spellings as very bad design.

PR #235, opened 2020-11-03
Adapter: Refactor response status

sonallux:
This PR fixes the second part of #156. Now meaningful HTTP status codes are returned from the REST endpoints of the Adapter.

### DIFF
--- a/adapter/integration-test/src/adapter-datasource-trigger.test.js
+++ b/adapter/integration-test/src/adapter-datasource-trigger.test.js
@@ -144,7 +144,15 @@ describe('Datasource triggering', () => {
     })
   })
 
-  test('Should publish results for failing datasources [POST /datasources/{id}/trigger]', async () => {
+  test('Should return 404 NOT_FOUND when trigger unknown datasources [POST /datasources/0/trigger]', async () => {
+    const triggerResponse = await request(ADAPTER_URL)
+      .post('/datasources/0/trigger')
+      .send()
+
+    expect(triggerResponse.status).toEqual(404)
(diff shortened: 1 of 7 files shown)


georg-schwarz (dismissed):

MBuchalik:
Is there any particular reason why the string "LOL" is being used here so often? Doesn't really sound professional to me, tbh ;) But I haven't yet worked with the ODS code, so there might be a simple explanation :)

(I saw that the string was being used before as well. So I am curious about the "why" :D)

sonallux:
> Is there any particular reason why the string "LOL" is being used here so often? Doesn't really sound professional to me, tbh ;) But I haven't yet worked with the ODS code, so there might be a simple explanation :)
> 
> (I saw that the string was being used before as well. So I am curious about the "why" :D)

`LOL` is actually only used in the test cases for validating that an unsupported input is rejected. But any other non-valid string could be used instead too. So I think there is no real reason why `LOL` is used.

I personally have just copied what others did before, I was just too lazy to come up with something new :D

MBuchalik:
Alright, thanks for the explanation :)

georg-schwarz:
Well, maybe we should use more serious terms.. 😅 Should we change it to something more professional?

Some suggestions: "NOT-SUPPORTED", "NULL", ...

sonallux:
I have replaced all the `LOLs` with something meaningful. This would also make @andreas-bauer happy that I am following the Boy Scout Rule 🤪

georg-schwarz (approved):

PR #242, opened 2020-11-12
Adapter: allow long URLs

### LINKED ISSUE #240: Adapter: Creating datasource with long url fails.
### Reproduction:

Try to create a new datasource using the Web-UI, use this url:

`https://services7.arcgis.com/mOBPykOjAyBO2ZKk/arcgis/rest/services/RKI_Landkreisdaten/FeatureServer/0/query?where=1%3D1&outFields=RS,GEN,BEZ,cases,deaths,cases_per_100k,cases_per_population,county,cases7_bl_per_100k,cases7_per_100k&returnGeometry=false&outSR=4326&f=json`

### Expected result:

works

### actual result:

no error message, no UI hint, network-tab on browser states: 
> Failed to load resource: the server responded with a status of 500 (Internal Server Error)
DatasourceCreate.vue?b91b:88 Failed to create datasource Error: Request failed with status code 500

adapter error log states:

> adapter-db_1            | 2020-11-11 09:26:39.934 UTC [71] ERROR:  value too long for type character varying(255)
adapter-db_1            | 2020-11-11 09:26:39.934 UTC [71] STATEMENT:  insert into datasource (format_parameters, format_type, author, creation_timestamp, description, display_name, license, protocol_parameters, protocol_type, first_execution, interval, periodic) values ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12)
adapter-db_1            | 	RETURNING *
adapter_1               | 2020-11-11 09:26:39.935  WARN 1 --- [nio-8080-exec-6] o.h.engine.jdbc.spi.SqlExceptionHelper   : SQL Error: 0, SQLState: 22001
adapter_1               | 2020-11-11 09:26:39.935 ERROR 1 --- [nio-8080-exec-6] o.h.engine.jdbc.spi.SqlExceptionHelper   : ERROR: value too long for type character varying(255)

### comment

We should really support URIs that are longer than 255 characters. Also we should offer an error message in this case nonetheless.

lunedis:
closes #240 .

The `protocol-parameters` are JSONified and then saved in the database. It looks like the standard length for a string field is 255, which I'd argue is too low, especially considering this field does not only contain the url (`location`) but also parameters that can be set inside the URL.

I changed the length to 2000, which is up for discussion.

Also we should add a validation and error message to it, but I honestly have no clue yet where one would add that, but I wanted this fixed for my Corona example. Maybe just open another issue for that? Feedback is appreciated.

### DIFF
--- a/adapter/integration-test/src/adapter-datasource.test.js
+++ b/adapter/integration-test/src/adapter-datasource.test.js
@@ -98,6 +98,18 @@ describe('Datasource Configuration', () => {
     expect(datasourceResponse.status).toEqual(400)
   }, TIMEOUT)
 
+  test('Should create datasource with long URL [POST /datasources]', async () => {
+    const longUrlDatasourceConfig = getDatasourceConfig()
+    const queryParameter = '&parameter=longParameter'
+    const longUrl = 'http://www.disresprect.com?first=test' + queryParameter.repeat(20)
+    longUrlDatasourceConfig.protocol.parameters.location = longUrl
+    const response = await request(ADAPTER_URL)
+      .post('/datasources')
+      .send(longUrlDatasourceConfig)
+
+    expect(response.status).toEqual(201)
+  })
+
   test('Should update existing datasource [PUT /datasources/{id}]', async () => {
     const postResponse = await request(ADAPTER_URL)
       .post('/datasources')
--- a/adapter/integration-test/src/adapter-datasource.test.js
+++ b/adapter/integration-test/src/adapter-datasource.test.js
@@ -100,8 +100,8 @@ describe('Datasource Configuration', () => {
 
   test('Should create datasource with long URL [POST /datasources]', async () => {
     const longUrlDatasourceConfig = getDatasourceConfig()
-    const queryParameter = '&veryLongProfessionalParameter=verylongProfessionalParameterValue'
-    const longUrl = 'http://www.very-long-professional-location-that-might-not-fit-in-the-database.com?first=test' + queryParameter.repeat(40)
+    const queryParameter = '&veryLongParameter=verylongParameterValue'
+    const longUrl = 'http://www.very-long-location-that-might-not-fit-in-the-database.com?first=test' + queryParameter.repeat(40)
     longUrlDatasourceConfig.protocol.parameters.location = longUrl
     const response = await request(ADAPTER_URL)
       .post('/datasources')
@@ -120,7 +120,7 @@ describe('Datasource Configuration', () => {
       .get('/datasources/' + createdDatasource.id)
 
     const updatedConfig = getDatasourceConfig()
-    updatedConfig.protocol.parameters.location = 'http://www.professional-location.com'
+    updatedConfig.protocol.parameters.location = 'http://www.location.com'
 
     const putResponse = await request(ADAPTER_URL)
       .put('/datasources/' + createdDatasource.id)
@@ -255,7 +255,7 @@ const getDatasourceConfig = () => ({
   protocol: {
     type: 'HTTP',
     parameters: {
-      location: 'http://www.professional-location.com'
+      location: 'http://www.location.com'
     }
   },
   format: {
@@ -268,7 +268,7 @@ const getDatasourceConfig = () => ({
     interval: 50000
   },
   metadata: {
-    author: 'professional-author',
+    author: 'author',
     license: 'none',
     displayName: 'test datasource 1',
     description: 'integration testing datasources'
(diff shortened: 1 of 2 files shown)


adapter/integration-test/src/adapter-datasource.test.js
```diff
@@ -98,6 +98,18 @@ describe('Datasource Configuration', () => {
     expect(datasourceResponse.status).toEqual(400)
   }, TIMEOUT)
 
+  test('Should create datasource with long URL [POST /datasources]', async () => {
+    const longUrlDatasourceConfig = getDatasourceConfig()
+    const queryParameter = '&parameter=longParameter'
+    const longUrl = 'http://www.disresprect.com?first=test' + queryParameter.repeat(20)
```
georg-schwarz: Please use other URL (seems not appropriate)

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/model/DatasourceProtocol.java
```diff
@@ -19,7 +19,7 @@
   private Protocol type;
 
   @NotNull
-  @Column(name = "protocol_parameters")
+  @Column(name = "protocol_parameters", length = 2000)
```
georg-schwarz: I think we should not use magic numbers here. Please extract to a constant.

georg-schwarz (changes requested):

sonallux:
> I changed the length to 2000, which is up for discussion.

Why not using a data type with unlimited length? We never know what and how many parameters users want to specify.

Adding the `@Lob` annotation to that field should make JPA use the unlimited `text` [data type of PostgreSQL](https://www.postgresql.org/docs/current/datatype-character.html).

adapter/integration-test/src/adapter-datasource.test.js
lunedis: done.

there are other occurrences of domains like that in the test case (not written by me), should i change those as well?

georg-schwarz: yep, please use the scouts rule => leave place a little cleaner than you found it :)

adapter/integration-test/src/adapter-datasource.test.js
```diff
@@ -98,6 +98,18 @@ describe('Datasource Configuration', () => {
...
+    const queryParameter = '&parameter=longParameter'
+    const longUrl = 'http://www.verylongdomain.com?first=test' + queryParameter.repeat(20)
+    longUrlDatasourceConfig.protocol.parameters.location = longUrl
+    const response = await request(ADAPTER_URL)
+      .post('/datasources')
+      .send(longUrlDatasourceConfig)
+
+    expect(response.status).toEqual(201)
+  })
+
```
mathiaszinnen, line 112: What happens if someone attempts to POST a datasource with an URL longer than 2000 characters? Does the service crash or does it respond with HTTP 500? 
Maybe we should add another test to ensure the correct behavior.

The cleanest solution would be to check the incoming request in the API class and respond with HTTP 400 if fields are too long.
However, since this probably applies to more fields, it should be done in another PR.

lunedis:
> > I changed the length to 2000, which is up for discussion.
> 
> Why not using a data type with unlimited length? We never know what and how many parameters users want to specify.
> 
> Adding the `@Lob` annotation to that field should make JPA use the unlimited `text` [data type of PostgreSQL](https://www.postgresql.org/docs/current/datatype-character.html).

Very good idea. I checked the PostgreSQL documentation (https://www.postgresql.org/docs/current/datatype-character.html) and it says

> There is no performance difference among these three types, apart from increased storage space when using the blank-padded type, and a few extra CPU cycles to check the length when storing into a length-constrained column. While character(n) has performance advantages in some other database systems, there is no such advantage in PostgreSQL; in fact character(n) is usually the slowest of the three because of its additional storage costs. In most situations text or character varying should be used instead.

So i opted to use the TEXT datatype.

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/model/DatasourceProtocol.java
lunedis: Resolved by not having a maximum length anymore.

adapter/integration-test/src/adapter-datasource.test.js
lunedis: > What happens if someone attempts to POST a datasource with an URL longer than 2000 characters? Does the service crash or does it respond with HTTP 500?
> Maybe we should add another test to ensure the correct behavior.

With the change to support unlimited length this is no longer an issue.

> However, since this probably applies to more fields, it should be done in another PR.

I agree we should add more validation logic in another pull request.

georg-schwarz (approved):

PR #243, opened 2020-11-13
Notification: Use @jvalue/node-dry-basics library instead of own implementations

georg-schwarz:


MBuchalik:
(Not a full review - just something I noticed.) The commit message seems to be broken: It says `Use @value/node-dry-basics` instead of `Use @jvalue/node-dry-basics`. The same also goes for the title of this PR.

georg-schwarz:
Thanks!
It's kind of strange... Linter says ok, but the library cannot be found at runtime. Probably I did something wrong in packaging the lib?

image

MBuchalik:
So it can also be compiled without any issues, but the library is missing during runtime? Or does the compiler also fail at some point?

Are there any logs during runtime, or is the library just `undefined` (or `null`)?

notification/package.json
```diff
@@ -43,7 +43,8 @@
     "jest": "^26.5.3",
     "ts-jest": "^26.4.1",
     "tslib": "^2.0.3",
-    "typescript": "^4.0.3"
+    "typescript": "^4.0.3",
+    "@jvalue/node-dry-basics": "0.0.3"
```
sonallux: You must move this from `devDependencies` to `dependencies`

sonallux (changes requested):

notification/package.json
georg-schwarz: oh darn, I'm so stupid... Thanks a lot <3

georg-schwarz:
Ah didn't fix the issue. That's the logs:
image

Does the published lib have to have the index.js at root?

MBuchalik:
Just to make sure: Did you run `npm install` after changing from `devDependencies` to `dependencies`?

georg-schwarz:
Jep :) I rebuilt the docker image without cache, so it should pull in as runtime dependency..

sonallux:
> Jep :) I rebuilt the docker image without cache, so it should pull in as runtime dependency..

@georg-schwarz  in your last commit 2cf851a where you moved the dependency from `devDependencies` to `dependencies` you have not committed the updated `package-lock.json`. This file should have been updated by `npm install`

When I do an `npm install` on your branch locally, rebuild the containers and run the notification integration tests, they all pass 🚀

georg-schwarz:
Phew, that was quite an odyssey... I guess I didn't trust my published npm package :)
Thanks @jsone-studios  and @MBuchalik for your help!

notification/src/env.ts
```diff
@@ -1,27 +1,15 @@
-const isEmpty = (value: string | undefined): value is undefined => value === undefined || value === ''
+import { readEnvOrDie } from '@jvalue/node-dry-basics'
```
mathiaszinnen, line 1: I really like that name

mathiaszinnen (approved):
Nice!

PR #246, opened 2020-11-14
Documentation: Update docker-compose instructions

### LINKED ISSUE #241: Adapter: README describing how to start integration test is wrong.
The README states:

> For integration testing run `docker-compose -f ../docker-compose.yml -f ../docker-compose.ci.yml up adapter-service adapter-service-it`

Obviously docker-compose.ci.yml doesn't exist anymore, and the service names changed, but changing the command into `docker-compose -f ../docker-compose.yml -f ../docker-compose.st.yml up adapter adapter-it` fails on my machine with

> WARNING: The DOCKER_REGISTRY variable is not set. Defaulting to a blank string.
> Building adapter
> ERROR: Couldn't connect to Docker daemon at http+docker://localhost - is it running?
> If it's at a non-standard location, specify the URL with the DOCKER_HOST environment variable.

If I go into the root directory and use `docker-compose -f docker-compose.yml -f docker-compose.it.yml up adapter adapter-it` it works fine.

Is that a setup error on my part? Do we want to change it to be run from the root directory?

mathiaszinnen:
closes #241. Additionally updates the docker-compose instructions in README files of the other services.

### DIFF
--- a/notification/README.md
+++ b/notification/README.md
@@ -16,9 +16,9 @@ Use `npm test` to run the unit tests.
 
 ## Running end-to-end tests
 
-Run integration tests in docker via the following command:
+Run integration tests in docker via the following command in the project root directory (open-data-service):
 
-```docker-compose -f docker-compose.yml -f docker-compose.it.yml build notification notification-it && docker-compose -f docker-compose.yml -f docker-compose.it.yml up notification notification-it edge```
+```docker-compose -f docker-compose.yml -f docker-compose.it.yml up --build notification notification-it edge```
 
 
 ## API
--- a/notification/README.md
+++ b/notification/README.md
@@ -18,7 +18,10 @@ Use `npm test` to run the unit tests.
 
 * For integration testing run 
   
-  ```docker-compose -f ../docker-compose.yml -f ../docker-compose.it.yml --env-file ../.env up notification notification-it```
+  ```docker-compose -f ../docker-compose.yml -f ../docker-compose.it.yml --env-file ../.env up notification-it```
+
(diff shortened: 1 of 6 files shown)


notification/README.md
```diff
@@ -16,9 +16,9 @@ Use `npm test` to run the unit tests.
 
 ## Running end-to-end tests
 
-Run integration tests in docker via the following command:
+Run integration tests in docker via the following command in the project root directory (open-data-service):
```
lunedis: any special reason why you recommend to start the notification-it from the root directory, but all other services from their own?

A quick test showed me that starting the notification-it from the subdirectory works as well (if you do it in similar style to the others)

lunedis (commented):
First of all, thanks, it works now with the added `--env-file` option.

Another thing: might be a bit out of scope, but what about adding the service-to-test to the "depends-on" clause of the relevant integration test?

This would result in the following:

- no need to specify both e.g. `adapter` and `adapter-it` when starting the test
- `docker-compose` doesn't show the output of the main service (can be both good or bad). Makes the output of the integration test cleaner, might be annoying if trying to debug though.

And one last thing: after running an integration test with docker-compose all services that those services depend on (e.g. `rabbit-mq`, the corresponding database, etc.) are still running. Do we want to add a note somewhere to remind people to stop those if they want?

notification/README.md
mathiaszinnen: > any special reason why you recommend to start the notification-it from the root directory, but all other services from their own?
> 
> A quick test showed me that starting the notification-it from the subdirectory works as well (if you do it in similar style to the others)

Good point. I will adjust the notification README to be consistent with the other services.

lunedis (approved):
LGTM

georg-schwarz (approved):

PR #253, opened 2020-11-23
Notification use @jvalue/node-dry-amqp

georg-schwarz:


lunedis (dismissed):
LGTM

georg-schwarz:
Just noticed that the consumer registering the method on the queue was not too clean. I refactored it by using explanatory variables, hope this makes it more explicit what which parameters mean!

lunedis (approved):
That was the only thing I disliked in the previous version of this PR, but not enough to disapprove.

But I like this version much more! Now I do not have to open the node-dry-amqp repository to understand the code.

georg-schwarz:
Yeah, you're welcome to also demand some style issues be fixed. 

Sometimes you just miss those small things during coding, but it is important to revisit and make it clean! It just appeared to me when having a last look before merging.

Those small things add up and will to much better code in the end.

Issue #238, opened 2020-11-10
Unify used docker images (both base and services)


lunedis:
A quick look over the Dockerfiles and the `docker-compose` file made me realize that we should make sure to use consistent versions of all used docker images.

We seem to be consistent in using `node:lts-alpine` for all node-related containers, but I'd like to double check.

We certainly use different versions of `postgres` (`latest` and `12-alpine`). Is there a reason for that?

Also do we even want to use something like `latest` or `lts-alpine` or rather specify an exact version? What's the best practice here?

sonallux:
IMO, it's best practice to **not** use the `latest` or `lts-alpine` docker tags and instead use a tag with a fixed major version number like `12-alpine`. The docker tags without a major version specified can move to a new major version number without notice. As nearly all projects do follow semantic versioning, such a change in the major version number does contain breaking changes and potentially break the ODS.

For example, the `node:lts-alpine` image we are using for all services has moved from node 12.X to node 14.X just two weeks ago.

georg-schwarz:
I agree we should depend on concrete major versions. We had the issue of things crashing due to version updates!
However, we have to make sure to check regularly for newer versions and upgrade manually. Building up technical depts in that regard is not a good idea as well! Any ideas how we want to handle that? Would it make sense to distribute responsibilities to services for maintaining?

MBuchalik:
> However, we have to make sure to check regularly for newer versions and upgrade manually.

Two ideas:
1. We could write a script that parses all dockerfiles and informs us about major releases. This script could be run manually, or using a cronjob, or... If a new major release is out, it could maybe even open a GitHub Issue. 

2. Use something like Dependabot. Seems like this is natively supported in GitHub: https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates
One thing I am not sure about:
Such tools often create a new PR for every single dependency update. No matter if it is a major release, or just a patch. So if you depend on something that changes really often, you will get lots of PRs. I don't know if it is possible to configure Dependabot to only inform us about major releases.

georg-schwarz:
Dependabot would be an option, however I think @dirkriehle mentioned that these PRs are legally not to be accepted? But we could use it to inform us somehow, I think it is configurable in some ways..

lunedis:
https://github.com/dependabot/dependabot-core/issues/390

It also seems that dependabot only supports image versions in Dockerfiles, not in `docker-compose` files.

[...]

PR #263, opened 2020-11-26
Cleanup traefik labels in docker-compose.yml

sonallux:
Currently, only some of our microservices are defining the `traefik.http.services.<service-name>.loadbalancer.server.port=<port>` label. According to the [Traefik documentation](https://doc.traefik.io/traefik/providers/docker/#port-detection), this label is not required when the container is exposing exactly one port. As all services expose only one port (see the `EXPOSE` statement in the `Dockerfile`s), I would remove this label everywhere.

The RabbitMQ management UI is currently accessible directly via `localhost:15672`. And there is a traefik route `/api/amqp` configured too, which I have fixed to use the correct port. 

But actually, I think that we should not expose the RabbitMQ Management UI under `/api/amqp` as it is only for administration purposes. Therefore I would like to remove exposing it under `/api/amqp`. What do you think? @georg-schwarz @mathiaszinnen @lunedis If you are ok with that I am going to update this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,9 +33,6 @@ This configuration can be done programmatically via the API or browser based wit
 ## Using the API
 There is a collection of examples for entire configurations in our [example request collection](doc/example-requests/).
 
-Additionally, there is is swagger API documentation available under `localhost:9400`. 
-The swagger integration is currently in progress, so there is only documentation for the storage service available.
-
 ## Using the UI
 
 The easiest way to use the ODS is via the UI. If you started the ODS with docker-compose you can access the UI under `http://localhost:9000/`.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,7 +19,6 @@ services:
     labels:
       - "traefik.enable=true"
       - "traefik.http.routers.to-ui.rule=PathPrefix(`/`)"
-      - "traefik.http.services.ui.loadbalancer.server.port=80"
 
   # ----------------- ADAPTER SERVICE (/adapter) --------------------
   adapter:
@@ -47,7 +46,6 @@ services:
       - "traefik.http.routers.to-adapter.rule=PathPrefix(`/api/adapter`)"
       - "traefik.http.routers.to-adapter.middlewares=adapter-stripprefix@docker"
       - "traefik.http.middlewares.adapter-stripprefix.stripprefix.prefixes=/api/adapter"
-      - "traefik.http.services.adapter.loadbalancer.server.port=8080"
 
   adapter-db:
     image: postgres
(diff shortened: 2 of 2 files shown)


georg-schwarz:
Thanks for removing the unnecessary clutter! 

For the management console, I don't see pressure removing it since it should be password protected. However, I don't think it is used at all, right? If so, we can also remove it I guess...

sonallux:
> However, I don't think it is used at all, right?

The RabbitMQ Managemnt UI is quite useful for debugging AMQP messages and manually sending messages. But I always use it directly via `localhost:15672`.

---

Another question. What is this service doing? It can not be accessed from the outside, because it does not have a `ports` mapping.
```
  storage-swagger: # API documentation for storage service
    image: swaggerapi/swagger-ui
    environment:
      API_URL: http://localhost:9400/
```

sonallux:
How about comment this service out, as it is just for managing Postgres? If anyone wants to use it he can uncomment it.
```
  storage-db-ui: # management UI for Postgres
    image: adminer
    ports:
      - 8080:8080
```

mathiaszinnen:
> The RabbitMQ management UI is currently accessible directly via `localhost:15672`. And there is a traefik route `/api/amqp` configured too, which I have fixed to use the correct port.
> 
> But actually, I think that we should not expose the RabbitMQ Management UI under `/api/amqp` as it is only for administration purposes. Therefore I would like to remove exposing it under `/api/amqp`. What do you think? @georg-schwarz @mathiaszinnen @lunedis If you are ok with that I am going to update this PR.

As long as we still expose it directly, we should remove the API route, yes. 

I think I introduced it for debugging, did not manage to make it work via traefik, switched to exposing the port directly and forgot to remove it afterwards. So thanks for finding it!

georg-schwarz:
Ok, so I think the RabbitMQ Management UI should have no use in production I guess, so we can remove that port in Traefik.

The swagger thing is deprecated and can be removed. Although, on the long run we want to have a fully working swagger-ui for the whole API.

The adminer can be commented out, but should also be renamed. It's just for debugging.

lunedis:
I'm currently working on #238. Good to know that you want to comment out certain services, then I don't have to handle those there.

sonallux:
I have updated my PR:
- Remove RabbitMQ Management UI from traefik. It is still accessible on `localhost:15672`
- Remove `storage-swagger` service
- Comment out `adminer` service

Now two fewer services have to be started 🚀

georg-schwarz (approved):

mathiaszinnen (approved):

PR #266, opened 2020-11-27
Adapter raw data

### LINKED ISSUE #262: Adapter: return raw data on request (raw interpreter)
Add a raw interpreter to return uninterpreted data. Required by workbench.

mathiaszinnen:
closes #262

### DIFF
--- a/adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/Adapter.java
+++ b/adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/Adapter.java
@@ -37,7 +37,7 @@ public class Adapter {
     try {
       var rawData = importer.fetch(config.protocolConfig.parameters);
       var result = interpreter.interpret(rawData, config.formatConfig.parameters);
-      return dataBlobRepository.save(new DataBlob(result.toString()));
+      return dataBlobRepository.save(new DataBlob(result));
     } catch (IOException e) {
       throw new IllegalArgumentException("Not able to parse data as format: " + config.formatConfig.format, e);
     }
--- a/adapter/README.md
+++ b/adapter/README.md
@@ -58,7 +58,7 @@ When started via docker-compose *base_url* is `http://localhost:9000/api/adapter
       }
     }
     "format": {
-      "type": "JSON" | "XML" | "CSV"
+      "type": "JSON" | "XML" | "CSV" | "RAW"
       "parameters": { } | CSVParameters
     }
   }
@@ -130,7 +130,7 @@ When started via docker-compose *base_url* is `http://localhost:9000/api/adapter
 ```
 {
   "format": {
-    "type": "JSON" | "XML" | "CSV"
+    "type": "JSON" | "XML" | "CSV" | "RAW"
     "parameters": { } | CSVParameters
   }
 }
--- a/adapter/integration-test/src/adapter-datasource.test.js
+++ b/adapter/integration-test/src/adapter-datasource.test.js
@@ -212,6 +212,36 @@ describe('Datasource Configuration', () => {
     expect(delResponse.status).toEqual(204)
   }, TIMEOUT)
 
+  test('Should not update datasource with raw format [PUT /datasources/{id}]', async () => {
+    const postResponse = await request(ADAPTER_URL)
+      .post('/datasources')
+      .send(getDatasourceConfig())
+    expect(postResponse.status).toEqual(201)
+    const datasourceId = postResponse.body.id
+
+    const originalGetResponse = await request(ADAPTER_URL)
+      .get('/datasources/' + datasourceId)
+    expect(originalGetResponse.status).toEqual(200)
+
+    const rawDatasourceConfig = getDatasourceConfig()
+    rawDatasourceConfig.format.type = 'RAW'
+    const datasourceResponse = await request(ADAPTER_URL)
+      .put('/datasources/' + datasourceId)
+      .send(rawDatasourceConfig)
+    expect(datasourceResponse.status).toEqual(400)
+
+    const updatedGetResponse = await request(ADAPTER_URL)
+      .get('/datasources/' + datasourceId)
+    expect(updatedGetResponse.status).toEqual(200)
+
+    expect(originalGetResponse.body).toEqual(updatedGetResponse.body)
+
+    const delResponse = await request(ADAPTER_URL)
+      .delete('/datasources/' + datasourceId)
+      .send()
+    expect(delResponse.status).toEqual(204)
+  }, TIMEOUT)
+
   test('Should delete specific datasource [DELETE /datasources/{id}]', async () => {
     const response = await request(ADAPTER_URL)
       .post('/datasources')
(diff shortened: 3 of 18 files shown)


georg-schwarz:
Please remove the unnecessary `toString()` in `Adapter.java` line 40.

georg-schwarz:
Question that is unrelated to your PR: Why do we store the data blob of a preview? Shouldn't that be part of the datasource workflow instead of the adapter?

mathiaszinnen:
> Question that is unrelated to your PR: Why do we store the data blob of a preview? Shouldn't that be part of the datasource workflow instead of the adapter?

Yeah, good point. I agree. I will create an issue for that

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/Adapter.java
```diff
@@ -37,7 +37,7 @@ public DataBlob executeJob(AdapterConfig config) throws IllegalArgumentException
     try {
       var rawData = importer.fetch(config.protocolConfig.parameters);
       var result = interpreter.interpret(rawData, config.formatConfig.parameters);
-      return dataBlobRepository.save(new DataBlob(result.toString()));
+      return dataBlobRepository.save(new DataBlob(result));
```
sonallux, line 40: Before your change, a DataBlob was always a stringified JSON object. Now it can be any string if the `RAW` interpreter is used. The DataBlob string is also present in the published import success event. Therefore all consumer of this event must now be able to handle any string and do not make the assumption that the string is a stringified JSON object. 

According to the dependency graph, the pipeline and the notification service are consuming the import success event. Have you checked what happens in those services if the `RAW` interpreter is used?

mathiaszinnen: I don't really understand what you mean. A `DataBlob` Object was and still is, marshalled as a JSON object.
Are you talking about the `data` attribute of the `DataBlob`? This was a JSON before and can be any string now. 
This feature was requested by @MBuchalik  for the data engineering workbench. 
I think it makes sense to provide a raw representation of the imported data to be able to choose how to interpret it later.

Are you worried about consumers not being able to interpret the `data` field of the `DataBlob`? I would argue that we can assume that, when consumers request a `RAW` import, they should be able to handle non-JSON data, but I don't know if I got your comment right...

sonallux: > Are you talking about the `data` attribute of the `DataBlob`?

Yes exactly. Sorry, what I was not clear enough.

> Are you worried about consumers not being able to interpret the `data` field of the `DataBlob`?

Yes exactly, especially the pipeline service, because I think it is currently assuming that the `data` field is a stringified JSON.
And we should not assume anything from our users, especially from a fault tolerance perspective. They can create pipelines for `RAW` interpreted data and therefore we must make sure the pipeline service does at least not crash because of that.

georg-schwarz: Hmm, I think what Jonas detected is a little gap in our domain.
For the adapter (/preview) we need the RAW interpreter, and for the datasource (-> triggering pipelines) we actually do not (yet).. That means we should still ensure that only JSON is sent to downstream services. 
Maybe an alternative implementation approach makes sense, where the AdapterEndpoint checks if a Format is specified, and if not the raw data is returned? Or even add another endpoint for that? (/dataImport/raw)

mathiaszinnen: <s>Okay, I think I get what you mean now. I agree that we should ensure fault tolerance but maybe there is a misunderstanding here:

when triggering a data import via the ``/dataImport`` endpoint, we do not publish an amqp event. This only happens at the ``/datasource/{id}/trigger`` endpoint so the pipeline service is not affected by this. 
 
Does this resolve your concerns or did I misunderstand you?</s>

Edit: Thinking about it again, I see the problem. The model for datasources changes as well, so it is possible to create a datasource with a RAW interpreter, which is then published after execution. 

Maybe the underlying problem is that the models for the stateless adapter and the stateful datasource package are coupled. So instead of adding another endpoint, we could create separated models for these two packages.. what do you think?

georg-schwarz: I think the shorter solution would be to add a second endpoint to just do the import without interpretation ;)

mathiaszinnen: I decided to check for RAW format in the datasource creation and update endpoints.
I think this is the shortest and cleanest solution because adding another endpoint would have implied a lot of redundant code or fundamental refactoring.

georg-schwarz:
I played around a little and came up with this version:
* refactor adapter to have separate methods for executing the whole adapter, the protocol, and the format.
* add new endpoint to only execute the protocol

You can see the diff here to this version: https://github.com/jvalue/open-data-service/compare/adapter-raw-data...adapter-raw-data-endpoint

IMHO no unnecessary redundant code and clean distinction between concepts. What do you think?

sonallux:
I like Georg's concept 👍

mathiaszinnen:
>  played around a little and came up with this version:
> 
>     * refactor adapter to have separate methods for executing the whole adapter, the protocol, and the format.
> 
>     * add new endpoint to only execute the protocol
> 
> 
> You can see the diff here to this version: [adapter-raw-data...adapter-raw-data-endpoint](https://github.com/jvalue/open-data-service/compare/adapter-raw-data...adapter-raw-data-endpoint)
> 
> IMHO no unnecessary redundant code and clean distinction between concepts. What do you think?

It looks quite similar to the first version I created adding another endpoint.
One key difference is that martin preferred to have the resulting data wrapped in some kind of json object, so I used the datablob for that. That is one point, where I had to introduce a lot of redundancy. The other points were handling of exceptions and the includeData parameter. I did not want the adapter which is not an endpoint class to throw response status exceptions. But probably it is ok to do so. 
So, since you both seem to prefer the separate endpoint solution, I will use that and wrap the raw data import with some kind of json representation (probably still a DataBlob).

sonallux:
One thing I do not have fully understood is the benefit of having an endpoint that just returns the raw open data. Wouldn't it be much easier to just directly request the open data without the ODS?  @MBuchalik @mathiaszinnen

georg-schwarz:
> One thing I do not have fully understood is the benefit of having an endpoint that just returns the raw open data. Wouldn't it be much easier to just directly request the open data without the ODS?  @MBuchalik @mathiaszinnen 

It's just for getting a preview of the data when configuring. We decide it to go over the backend service instead of doing it in the UI directly to prevent code duplication and ensure consistent behavior.

In the operation of the ODS where the open data is piped through the ETL process this feature is not relevant.

MBuchalik:
Correct. There are basically two options we have: 
1. Use the ODS to fetch the raw data
2. Create a separate server just for this purpose

(Directly requesting data from the source via Ajax obviously won't work because of potentially missing CORS headers.)

Number 2 would, from my point of view, probably be an overkill for this small requirement.

In the future, if we figure out that there are more such requirements, then we could probably think about creating a separate backend for the workbench. But currently, we decided to use the ODS as the only backend for the workbench.

sonallux:
I missed that directly requesting the open data can cause CORS issue. Now everything makes sense. Thank you very much for the hint @MBuchalik

mathiaszinnen:
Continuing in #269

PR #269, opened 2020-12-03
Adapter raw data endpoint

mathiaszinnen:


adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/api/rest/v1/AdapterEndpoint.java
```diff
@@ -16,32 +16,58 @@
 @RequestMapping("/")
 public class AdapterEndpoint {
 
-  private final Adapter adapter;
+    private final Adapter adapter;
```
sonallux: Why did you change the indentation to 4 spaces? According to the `.editorconfig` file in the root of the repository, all files should be intended with two spaces.

sonallux (changes requested):

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/api/rest/v1/AdapterEndpoint.java
mathiaszinnen: I blame IntelliJ!
Thanks for pointing it out, I will fix that

adapter/README.md
```diff
@@ -50,20 +52,25 @@ When started via docker-compose *base_url* is `http://localhost:9000/api/adapter
...
-      "type": "HTTP",
-      "parameters": {
-        "location": String,
-        "encoding": String
-      }
-    }
+    "protocol": ProtocolConfig 
     "format": {
-      "type": "JSON" | "XML" | "CSV"
+      "type": "JSON" | "XML" | "CSV" | "RAW"
```
georg-schwarz: We removed that right?

adapter/README.md
```diff
@@ -130,7 +138,7 @@ When started via docker-compose *base_url* is `http://localhost:9000/api/adapter
 ```
 {
   "format": {
-    "type": "JSON" | "XML" | "CSV"
+    "type": "JSON" | "XML" | "CSV" | "RAW"
```
georg-schwarz: Same here

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/api/rest/v1/AdapterEndpoint.java
```diff
@@ -23,23 +23,49 @@ public AdapterEndpoint(Adapter adapter) {
...
+      throw new ResponseStatusException(HttpStatus.INTERNAL_SERVER_ERROR, e.getMessage());
+    }
+  }
+
+  @PostMapping(value = Mappings.RAW_PREVIEW_PATH, produces = "application/json")
+  public ResponseEntity<?> executeRawPreview(
+    @Valid @RequestBody ProtocolConfig config,
+    @RequestParam(required = false, defaultValue = "true") boolean includeData) {
+    try {
+      DataBlob imported = adapter.executeRawImport(config, includeData);
```
georg-schwarz: 2nd parameter here called `ìncludeData` and in method called `persistData`. Aren't those two different things? Are we mixing things up? Or is it two names for the same thing?

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/interpreter/CsvInterpreter.java
```diff
@@ -57,12 +57,12 @@ protected void validateParameters(Map<String, Object> inputParameters) {
   }
 
   @Override
-  protected JsonNode doInterpret(String data, Map<String, Object> parameters) throws IOException {
+  protected String doInterpret(String data, Map<String, Object> parameters) throws IOException {
```
georg-schwarz: Can we undo this change since we have no raw interpreter anymore?

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/api/rest/v1/DatasourceEndpoint.java
```diff
@@ -88,10 +88,8 @@ public void deleteAllDatasources() {
       .orElseThrow(() -> new ResponseStatusException(HttpStatus.NOT_FOUND, "Can not find Datasource with id: " + id));
     try {
       return datasourceManager.trigger(id, runtimeParameters);
-    } catch (IllegalArgumentException e) {
-      throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Invalid datasource or parameters: " + e.getMessage());
-    } catch (RestClientException e) {
-      throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Failed to load data: " + e.getMessage());
+    } catch (ResponseStatusException e) {
```
georg-schwarz, line 89: unnecessary

georg-schwarz (changes requested):

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/api/rest/v1/AdapterEndpoint.java
mathiaszinnen: IncludeData implies that it does not need to persisted. However, I should probably remove that parameter altogether since in a non-raw import requests with includeData param are persisted as well. 
Maybe we should prevent persistence in both cases, but this should probably be done in another PR, what do you think?

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/interpreter/CsvInterpreter.java
mathiaszinnen: Yeah, good point, I will do that. The JsonNode return type was more precise.

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/api/rest/v1/AdapterEndpoint.java
georg-schwarz: Yeah move to other PR. Maybe remove that parameter for the raw preview endpoint completly in this PR?

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/api/rest/v1/DatasourceEndpoint.java
mathiaszinnen: This is actually needed to prevent all response status exceptions from being caught by the subsequent catch clause.

georg-schwarz: Ok thanks, didn't see that in the diff..

georg-schwarz (approved):
LGTM

PR #271, opened 2020-12-04
fix(ui): avoid range error when setting first execution date

Kexplx:
`sliceMinuteFromTimeString` and `sliceHourFromTimeString` expect a time string in _HH:mm_ format  e.g. "12:01".
`getISOTimeString` generates a time string in _H:m_ format e.g. "12:1".

This leads to a RangeError when clicking "Ok" without changing the default time (returned by `getISOTimeString` and set during component load) in the time-picker within the datasource form.

**This PR**
1. Changes `getISOTimeString` to return the time in the actual ISO format "HH:mm"
2. Removes slice methods for date and time strings. A date can just be initialized like so: `new Date(this.date + ' ' + this.time) `
3. Moves the getISOString methods to helper file for potential reuse
4. Adds tests to validate correct format


lunedis:
Thanks for adding the tests! I just noticed that we don't actually run any UI tests on building or in the CI. I'm gonna open a new issue for that.

mathiaszinnen (dismissed):
LGTM

lunedis:
@Kexplx With #272 merged, please rebase this PR off master and move the tests to where you want them to be. The `tests`-folder is currently not considered anymore.

lunedis (changes requested):
Move `date-helpers-spec.ts` next to `date-helpers.ts`.

Kexplx:
Thanks for the reviews @lunedis and @mathiaszinnen. I've moved helpers & tests into `/src/helpers`.

Git history looks good aswell after rebasing.

Unbenannt

lunedis (dismissed):
LGTM

georg-schwarz:
LGTM.

Should we name tests as `*.spec.ts` or `*.test.ts`?
Since we are having many students having no particular web background, I'd rather go with `test` since it is easier to understand what is happening here..

Kexplx:
Having consistent file endings is definitely a good idea and `*.test.ts` also works for me.

georg-schwarz:
Nice, thanks @Kexplx !

georg-schwarz (approved):

PR #273, opened 2020-12-04
UI: run unit tests on build

### LINKED ISSUE #272: Run UI tests on build and CI
Merging #271 would add the first unit tests in our vue.js UI. 

We currently do not run those UI tests in the build `Dockerfile` or the CI.

lunedis:
Fixes #272 

I had to upgrade the version of `jest-html-reporter` because of https://github.com/Hargne/jest-html-reporter/issues/104.
Not sure if we even the output though... Should we just remove it?

Also I created the `package-lock.json` with my local Node v15 installation. Lets see if this breaks the CI...


lunedis:
> Also I created the `package-lock.json` with my local Node v15 installation. Lets see if this breaks the CI...

It doesn't! Yay.

Kexplx:
Looks good. I didn't even know we have that html generator installed. Are you guys using it?

georg-schwarz:
> Looks good. I didn't even know we have that html generator installed. Are you guys using it?

Good point. We are not using it... @lunedis would you mind removing it in this PR?

georg-schwarz (dismissed):

Kexplx:
Currently, jest only looks for test files inside of `ui/tests/unit`. I think it would be better to just put the tests right next to the code under test. That's also how it's done in the node-based, backend-parts of the ODS.

I suggest we add `'**/*.(spec|test).(js|jsx|ts|tsx)|**/__tests__/*.(js|jsx|ts|tsx)'` to the testMatch property in jest.config.ts and delete the `/tests` folder entirely.

What do you think?

mathiaszinnen:
> Currently, jest only looks for test files inside of `ui/tests/unit`. I think it would be better to just put the tests right next to the code under test. That's also how it's done in the backend-parts of the ODS.
> 
> I suggest we add `'**/*.(spec|test).(js|jsx|ts|tsx)|**/__tests__/*.(js|jsx|ts|tsx)'` to the testMatch property in jest.config.ts and delete the `/tests` folder entirely.
> 
> What do you think?

Good point! I would also prefer to keep it consistent. The single test in `ui/tests/unit` is just a placeholder anyway.

edit: I just saw you added some tests in your other PR, nice!

lunedis:
> Currently, jest only looks for test files inside of `ui/tests/unit`. I think it would be better to just put the tests right next to the code under test. That's also how it's done in the node-based, backend-parts of the ODS.
> 
> I suggest we add `'**/*.(spec|test).(js|jsx|ts|tsx)|**/__tests__/*.(js|jsx|ts|tsx)'` to the testMatch property in jest.config.ts and delete the `/tests` folder entirely.
> 
> What do you think?

I agree. I updated the PR accordingly.

I also changed the `package.json` to use `jest` directly, instead of the `vue-cli-service`. The CLI service defaults to using watch-mode, which is not suited for the docker workflow.

I also made sure `jest` does not fail if no tests are present.

mathiaszinnen (approved):

PR #279, opened 2020-12-07
system test: wait on storage-mq container

### LINKED ISSUE #278: Wait for StorageMQ in system test
The system test is waiting on startup for all microservices to be ready except the StorageMQ service. We should definitely wait for the StorageMQ service to be ready too.

lunedis:
closes #278

Pretty straightforward.


mathiaszinnen (approved):
Nice find, thanks for fixing

sonallux (approved):

PR #286, opened 2020-12-11
ui: update datasource table

Kexplx:
Straight forward changes to the datasource table inside `DataSourceOverview.vue`.

- Columns for `location` and `firstExecution` properties
- Timestamp conversion helpers moved into /helpers (Following #271 theme)
- Obvious style changes to the table layout

<hr>
Before

chrome_ovjmfAwDDD

After
chrome_AVbfIToZ9B

### DIFF
--- a/ui/src/App.vue
+++ b/ui/src/App.vue
@@ -77,13 +77,17 @@ export default class App extends Vue {
 </script>
 
 <style>
+html {
+  overflow: auto !important;
+}
+
 #app {
   font-family: 'Avenir', Helvetica, Arial, sans-serif;
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
-  text-align: center;
   color: #2c3e50;
 }
+
 #nav {
   padding: 30px;
 }
--- a/ui/src/datasource/DatasourceOverview.vue
+++ b/ui/src/datasource/DatasourceOverview.vue
@@ -1,150 +1,153 @@
 <template>
   <div class="datasource">
-    <v-card>
-      <v-card-title>
-        <v-btn
-          class="ma-2"
-          color="success"
-          @click="onCreate()"
-        >
-          Create new Datasource
-          <v-icon
-            dark
-            right
-          >
-            mdi mdi-pipe
-          </v-icon>
-        </v-btn>
-        <v-btn
-          class="ma-2"
-          @click="loadDataSources()"
-        >
-          <v-icon dark>
-            mdi mdi-sync
-          </v-icon>
-        </v-btn>
-        <v-spacer />
-        <v-text-field
-          v-model="search"
-          label="Search"
-          append-icon="mdi mdi-magnify"
-          single-line
-          hide-details
-        />
-      </v-card-title>
-
-      <v-data-table
-        :headers="headers"
-        :items="datasources"
-        :search="search"
-        :custom-filter="filterOnlyDisplayName"
-        :loading="isLoadingDatasources"
-        class="elevation-1"
+    <header class="d-flex mb-5">
+      <h1>Datasources</h1>
+      <v-spacer />
+      <v-btn
+        depressed
+        @click="loadDataSources()"
       >
-        <v-progress-linear
-          slot="progress"
-          indeterminate
+        <v-icon>
+          mdi-sync
+        </v-icon>
+      </v-btn>
+      <v-btn
+        class="ml-2"
+        color="primary"
+        @click="onCreate()"
+      >
+        Create Datasource
+      </v-btn>
+    </header>
+    <v-text-field
+      v-model="search"
+      class="mb-4"
+      label="Search name"
+      prepend-icon="mdi-magnify"
+      hide-details
+    />
+
+    <v-data-table
+      :headers="headers"
+      :items="datasources"
+      :single-expand="true"
+      :search="search"
+      :custom-filter="filterOnlyDisplayName"
+      :loading="isLoadingDatasources"
+      show-expand
+      class="elevation-3"
+    >
+      <v-progress-linear
+        slot="progress"
+        indeterminate
+      />
+
+      <template #[`item.trigger.periodic`]="{ item }">
+        <v-simple-checkbox
+          v-model="item.trigger.periodic"
+          disabled
         />
+      </template>
 
-        <template #[`item.trigger.interval`]="{ item }">
-          {{ getHoursFromMS(item.trigger.interval) }}h:{{ getMinutesFromMS(item.trigger.interval) }}m
-        </template>
+      <template #[`item.protocol.parameters.location`]="{ item }">
+        {{ trimQueryParams(item.protocol.parameters.location) }}
+      </template>
 
-        <template #[`item.trigger.periodic`]="{ item }">
-          <v-switch
-            v-model="item.trigger.periodic"
-            class="ma-2"
-            disabled
-          />
-        </template>
+      <template #[`item.action`]="{ item }">
+        <v-tooltip top>
+          <template #activator="{ on, attrs }">
+            <v-icon
+              v-bind="attrs"
+              small
+              class="mr-2"
+              v-on="on"
+              @click="onCreatePipeline(item)"
+            >
+              mdi-pipe
+            </v-icon>
+          </template>
+          <span>Create pipeline</span>
+        </v-tooltip>
+        <v-tooltip top>
+          <template #activator="{ on, attrs }">
+            <v-icon
+              v-bind="attrs"
+              small
+              class="mr-2"
+              v-on="on"
+              @click="onEdit(item)"
+            >
+              mdi-pencil
+            </v-icon>
+          </template>
+          <span>Edit</span>
+        </v-tooltip>
+        <v-tooltip top>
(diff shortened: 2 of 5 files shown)


ui/src/datasource/DatasourceOverview.vue
```diff
@@ -47,60 +40,74 @@
...
+            v-if="isPastDate(item.trigger.firstExecution)"
+            small
+          >
+            mdi-check
+          </v-icon>
+          <v-icon
+            v-else
+            small
+          >
+            mdi-clock-outline
```
georg-schwarz: Nice idea, but probably not worth to add that little icon.

I wonder which information here is relevant to the user... the first execution time like configured, or the last execution, or even the next planned execution? Probably doing more than one consumes too much space...

What do you think about showing the last execution date instead of the first one, and add this very icon here to the periodic column with a tooltip showing the next execution time?

ui/src/datasource/DatasourceOverview.vue
```diff
@@ -47,60 +40,74 @@
...
-          >
-            Create Pipeline
-            <v-icon
-              dark
-              right
-            >
-              mdi mdi-pipe
-            </v-icon>
-          </v-btn>
+          <v-tooltip bottom>
```
georg-schwarz: Very nice idea! Having it placed in the button seems like a waste of time.

ui/src/datasource/DatasourceOverview.vue
```diff
@@ -110,41 +117,55 @@
...
+} from '../helpers/date-helpers'
 import Datasource from './datasource'
 import * as DatasourceREST from './datasourceRest'
 
-const ONE_HOUR_IN_MS = 3600 * 1000
-const ONE_MINUTE_IN_MS = 60 * 1000
-
-@Component({})
+@Component({
+  methods: { getISOTimeString, getISODateString }
```
georg-schwarz: Interesting! Never saw it like that. How is it different from just importing and using it?

ui/src/datasource/DatasourceOverview.vue
georg-schwarz: We might also think about adding a `details` component in the future that shows all configuration information, including dependent pipelines (but definitely not part of this PR)

georg-schwarz (commented):
Nice work!

All in all it is mergable, but I would like to discuss if there is a better way to present the execution date-time information to the user (see inline comments).

Thanks for your contribution :)

ui/src/datasource/DatasourceOverview.vue
Kexplx: Showing `lastExecutionTime` and `nextExecutionTime` makes sense. We just have to be aware that both dates might not exist (yet).

> We might also think about adding a `details` component in the future that shows all configuration information, including dependent pipelines (but definitely not part of this PR)

Maybe we should even consider to split the table up into periodic and non-periodic datasources. That way, we can avoid all of the conditional logic inside the template and avoid ugly white columns.

Something else i noticed, showing the interval when the datasource isn't periodic doesn't make sense right? Currently it shows either way.

ui/src/datasource/DatasourceOverview.vue
Kexplx: This way, we can call the helpers directly from the template without having an additional method in the component class that just forwards the call.

ui/src/datasource/DatasourceOverview.vue
Kexplx: For some reason, the tooltips are displayed above the table when i remove the `bottom` directive. Looks very buggy, i'll investigate on this.

Kexplx: Simply had to replace `bottom` with `top` which i thought was the default.

image

ui/src/datasource/DatasourceOverview.vue
Kexplx: What do you think of this: 

image

Would be a pragmatic solution until we have a properly implemented `details` component.

georg-schwarz: I think that is a good pragmatic solution 👍

lunedis:
Can we improve the handling of very long URLs?

I just tried the URL used in the corona-check example, and it looks like this:

Screenshot_2020-12-14_09-45-07

Kexplx:
> I just tried the URL used in the corona-check example, and it looks like this:

Yeah that looks horrible, i'll look into it. Thanks for the review.

Kexplx:
> Can we improve the handling of very long URLs?

So a possible solution would do two things.

1. Make the columns break on "non-breakable" strings like URLs
2. Cut off query params and append '...' at the end

1 would break the line but we'd still be left with this enormous URL which takes up the space of multiple rows therefore i'd also do 2.

image

What do you think?

georg-schwarz:
Yeah, it should only be for a first glance. The details are exposed on the dropdown then :+1:

Kexplx:
What's up with the CI?

ui/src/App.vue
```diff
@@ -77,11 +77,14 @@ export default class App extends Vue {
...
 <style>
+html {
+  overflow: auto !important;
+}
+
 #app {
   font-family: 'Avenir', Helvetica, Arial, sans-serif;
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
-  text-align: center;
```
lunedis, line 84: Probably quite nitpicky, but I see you removed this rule to stop the text in the JSON editor to be centered.

Now the menu is not centered anymore (quite minimal change, but i actually liked the centered menu).

ui/src/datasource/DatasourceOverview.vue
```diff
@@ -154,29 +175,64 @@ export default class DatsourceOverview extends Vue {
...
+.editor {
+  width: 100%;
+  height: 500px;
+}
+
+.td-expanded {
+  padding: 0 !important;
+}
+
+.header{
```
lunedis: missing space

ui/src/datasource/DatasourceOverview.vue
```diff
@@ -154,29 +175,64 @@ export default class DatsourceOverview extends Vue {
...
 </script>
+<style>
+td {
+  word-break: break-all !important;
+}
+
+.editor {
+  width: 100%;
+  height: 500px;
+}
```
lunedis: I wonder if there is a way to make the editor always the required height without specifying a pixel value.

As it is right now, "normal" configs fit in perfectly (on my screen/machine at least), configs with a longer URL produce a scroll bar for 1 line.

lunedis (commented):
I only have 3 minor nitpicks.

If you are getting bored of them @Kexplx we can also just merge it in my opinion, especially because the JSON editor display is not meant to be the final solution anyways.

Despite all that, I kinda dislike seeing the CSS rules with `!important` that often, but I guess it's unavoidable when trying to adjust the UI library we use to our preferences.

ui/src/App.vue
Kexplx: I actually intended to left align the menu items and headings here. The editor's style was just a byproduct of the cascading css.

The convention, when looking at [Vuetify](https://vuetifyjs.com/en/components/), [Vue Docs](https://vuejs.org/v2/guide/), [Angular Docs](https://angular.io/start), is to align the menu items and headings on the left. It's completly up to you but conventions are always something valuable when it comes to web-usability.

ui/src/datasource/DatasourceOverview.vue
Kexplx: yep

ui/src/datasource/DatasourceOverview.vue
Kexplx: Agreed, i did some prior research on this without much success but i can look into it further.

Kexplx:
> I only have 3 minor nitpicks.

No problem. Thanks for the nitpicks @lunedis

ui/src/App.vue
lunedis: I agree, then I guess the way we did it before was unconventional. Let's stick to the left alignment then.

ui/src/datasource/DatasourceOverview.vue
Kexplx: I'm wondering if the editor was an overkill here. Since no editing will be done, the configuration overhead probably isn't justified. I think it's enough to simply wrap the JSON in `<pre>...</pre>`.

image

lunedis: I wondered about that, too, and I think i prefer the simple version with the `<pre>`-tag.

Kexplx:
Ready to merge. Thanks for the feedback @lunedis @georg-schwarz

lunedis (changes requested):
I'm actually starting to feel bad now, but when you use a long URL, and then expand the row, the `<pre>` field expands into the full width, which also makes the table wider, which hides the icons and the ability to collapse the row again. You need to scroll all the way to the left to do the operations.

Could you fix that?

Screenshot_2020-12-17_12-35-59

Kexplx:
> Could you fix that?

Done

lunedis (approved):
LGTM!

Thanks for the work and the unrelenting fixes over and over.

Issue #291, opened 2020-12-15
Adapter returns stringified array


MBuchalik:
When requesting data using the Adapter, we get a response like this:

```json
{"id":1234,"data":"[{\"column1\":\"first\",\"column2\":\"second\",\"column3\":\"third\"}]"}
```

# Expected result:

The "data" field should be an actual array.

```json
{
  "id":1234,
  "data": [
    {"column1":"first","column2":"second","column3":"third"}
  ]
}
```

# Actual result

The "data" field is currently a string. So you need to run `JSON.parse()` in order to get the actual data.

sonallux:
What exact endpoint are you using?

MBuchalik:
I am using the "preview" endpoint (`api/adapter/preview`). But it seems like it also happens with `api/adapter/dataImport?includeData=true`

sonallux:
Well, this is currently somehow by design (our internal events use this format too) and also because the Adapter is written in Java. As Java and JPA do not natively support JSON, we are storing the `data` as `String` internally, which is not ideal but simple.

Might be worth keeping this in mind when we are discussing #283 in the next meeting.

mathiaszinnen:
Exactly, we stored the data as a string because this was the simplest solution regarding persistence. 
I also think that we should discuss this in the developer meeting. 
However, I think we should keep it separate from #283.

PR #294, opened 2020-12-17
Adapter remove persistence

### LINKED ISSUE #267: Adapter: Imported data persistence
Currently, imported data is persisted in the stateless part of the adapter service (adapter package). This should be moved to the stateful part (datasource package).

mathiaszinnen:
closes #267

### DIFF
--- a/adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/api/rest/v1/Mappings.java
+++ b/adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/api/rest/v1/Mappings.java
@@ -2,9 +2,8 @@ package org.jvalue.ods.adapterservice.adapter.api.rest.v1;
 
 public class Mappings {
   public static final String DATA_PATH = "/data";
-  public static final String PREVIEW_PATH = "/preview";
   public static final String IMPORT_PATH = "/dataImport";
-  public static final String RAW_PREVIEW_PATH = "/preview/raw";
+  public static final String RAW_IMPORT_PATH = "/dataImport/raw";
   public static final String FORMAT_PATH = "/formats";
   public static final String PROTOCOL_PATH = "/protocols";
   public static final String VERSION_PATH = "/version";
--- a/adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/api/rest/v1/Mappings.java
+++ b/adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/api/rest/v1/Mappings.java
@@ -1,7 +1,6 @@
 package org.jvalue.ods.adapterservice.adapter.api.rest.v1;
 
 public class Mappings {
-  public static final String DATA_PATH = "/data";
   public static final String IMPORT_PATH = "/dataImport";
   public static final String RAW_IMPORT_PATH = "/dataImport/raw";
   public static final String FORMAT_PATH = "/formats";
--- a/adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/api/rest/v1/Mappings.java
+++ b/adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/api/rest/v1/Mappings.java
@@ -1,8 +1,8 @@
 package org.jvalue.ods.adapterservice.adapter.api.rest.v1;
 
 public class Mappings {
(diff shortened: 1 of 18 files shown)


adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/api/rest/v1/Mappings.java
```diff
@@ -1,10 +1,8 @@
 package org.jvalue.ods.adapterservice.adapter.api.rest.v1;
 
 public class Mappings {
-  public static final String DATA_PATH = "/data";
-  public static final String PREVIEW_PATH = "/preview";
   public static final String IMPORT_PATH = "/dataImport";
-  public static final String RAW_PREVIEW_PATH = "/preview/raw";
+  public static final String RAW_IMPORT_PATH = "/dataImport/raw";
```
mathiaszinnen: No reason to distinguish between preview and import anymore since everything is stateless now

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/api/rest/v1/Mappings.java
```diff
@@ -0,0 +1,5 @@
+package org.jvalue.ods.adapterservice.datasource.api.rest.v1;
+
+public class Mappings {
+  public static final String DATA_PATH = "/data";
```
sonallux, line 4: Maybe add the path `/datasources` from the `DatasourceEndpoint` here to, so all path mappings are defined in one place.

mathiaszinnen: good point!

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/api/rest/v1/Mappings.java
georg-schwarz: You're right. I think the name preview is more precisely though to indicate no downstream processing.

mathiaszinnen: I agree, preview fits better. We should think about changing the method type to GET as well. However, I think this should be done in another pr

georg-schwarz: I think changing to GET won't do since we are sending a body.

georg-schwarz:
We should also change the route in the UI for preview. Otherwise, the UI will break!

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/api/rest/v1/Mappings.java
mathiaszinnen: Good point, I did not think about that

mathiaszinnen:
> We should also change the route in the UI for preview. Otherwise, the UI will break!

Fortunately, the UI uses the `/datasource/{id}/trigger` endpoint

lunedis (commented):
In #291 you said (https://github.com/jvalue/open-data-service/issues/291#issuecomment-746836374) that you want to keep the JSON-as-string thing seperate from this issue.

Is this the reason why we return a JSON with `{ "data": "XXX" }` where XXX is either raw or JSON, compared to simply returning without wrapping it?

Other than that, looks good to me.

mathiaszinnen:
> In #291 you said ([#291 (comment)](https://github.com/jvalue/open-data-service/issues/291#issuecomment-746836374)) that you want to keep the JSON-as-string thing seperate from this issue.
> 
> Is this the reason why we return a JSON with `{ "data": "XXX" }` where XXX is either raw or JSON, compared to simply returning without wrapping it?

The wrapping in a JSON was a requirement that @MBuchalik had. 
I also think it makes sense to wrap it in a JSON. Otherwise, the consumer would not know what data format to expect in the response. Wrapped like this, we can guarantee a JSON response with one string attribute.

lunedis (commented):
Apart from #291 (which we agreed to handle seperately), looks good to me.

lunedis (approved):

PR #298, opened 2020-12-21
Fix scheduler initialization

### LINKED ISSUE #284: Scheduler can get initialized with wrong state
### Short overview
The scheduler stores a `SchedulingJob` for each datasource in memory. All changes to datasources (addition, update and deletion) are propagated to the scheduler by AMQP messages. The corresponding `SchedulingJob` is then adjusted accordingly, so the scheduler is always triggering the right datasource at the right time. The state of the scheduler is not persistent and therefore it must be reconstructed on every start of the scheduler. Because past datasource change events can not be replayed, the scheduler must query the adapter for the current list of datasources when it is starting. So basically the scheduler must perform two tasks when starting:
1. It must get the current list of datasources from the adapter and create the corresponding `SchedulingJobs`
2. It must start listening for changes to the datasources

If those two tasks are not perfectly aligned to each other, the following problems can occur which can lead to a wrong state in the scheduler:
- Changes can get lost
- Changes are applied in the wrong order
- Changes can get applied twice (Fortunately this does not cause issues, because applying a change in the scheduler is idempotent)

### Current scheduler implementation
Currently, the scheduler first starts listening to the datasource change events and then performs a REST request to the adapter to get the current list of all datasources. But this can lead to changes applied in the wrong order.

This happens because an event consumption can overtake the REST response delivery and creating the initial state. Then the changes are applied in the wrong order leading to a wrong state. There is a schematic visualization of this behavior in the following image. The black arrows are datasource changes, which trigger the datasource change events (green arrows). The blue arrows are the requests for querying the current list of datasources and the corresponding responses. 
Scheduler-init-1

<details><summary><strong>Non-working solution approach</strong></summary>
<p>
One might think this can be solved by changing the order of getting the current state and starting to consume events. But this can cause lost changes:

Scheduler-init-3

</p>
</details>

### Requirements for a working solution
To get a solution that guarantees consistency is very difficult and complex, so here are the main requirements for a solution:
- All changes to datasources starting at the point when the **adapter** constructed the list of datasources for the response, must arrive at the scheduler
- Events published after the **adapter** has created the current list of datasources, must not be applied before the current list of datasources has been applied

### Simple solution that does not have strong consistency
The following solution can not guarantee strong consistency and in my opinion not even eventual consistency. But that's discussable because at some point it will get to a consistent state:
1. Start listening for datasource change events, but do not yet apply the changes to the scheduler state
2. Request the current state from the adapter
3. Build the initial scheduler state
4. Start applying the changes from events

This works because the order is ensured by waiting and applying a change is idempotent. But this solution does not provide strong consistency. See the following image for an example. After applying the current list of datasources the scheduler state is consistent to the state of the adapter. Then the change of the first event is applied, which can lead to an inconsistent state. Only after applying the second event, the state is consistent again. I would argue this is not even eventual consistent, because an inconsistent state archived after being in a consistent state. The state will become consistent again because we can assume events do not get lost and all event will get consumed at some point in time.
Scheduler-init-5

### Other solutions
- Send the current state via an event. Ordering is guaranteed by RabbitMQ, if the messages are published on one channel. So the Adapter must ensure that Datasource change events and the current state event is published on the channel in the right order. The Scheduler would then start applying changes from events after consuming the initial state event.
- The Adapter constructs some kind of timestamp for the current state and sends this timestamp with the current state. This timestamp can be used to filter events and only apply events, that have been published after the current state has been created. RabbitMQ has no such feature, but other message/event system might have.
- Discard RabbitMQ and use Kafka so we can build a truly event-sourced system

### The important role of the adapter database
What I have not mentioned till now is that the adapter does not have the current state immediately available, it must always perform a network request to query the adapter-database. And as soon as the database has performed the query the result might be already outdated because another request could have already changed any datasource. The change event for later change can actually be delivered to the scheduler before the result of the first query is delivered to the scheduler. 

This is actually caused by another broader issue the current architecture has for all database, for which data change events are published: The order of the change events might not be equal to the order, in which the changes have been performed by the database. Please see my other issue for more details: #285.

### Conclusion
A nice solution can only be found when fixing the database event ordering issue (#285) first because the current issue is highly dependent on how and when events are published.
For the time being, we could use my simple solution that does hot have strong consistency. But nevertheless, this issue with the initialization of the scheduler must be reevaluated after the database event ordering issue has been fixed.

---

Wow, this got quite long 👀  I hope, I have not lost everyone somewhere in the unreliable network 😆 
If there is anything unclear or I have forgotten something please let me know.

sonallux:
Fixes #284

This copies some snippets of `node-dry-amqp` back to the scheduler because the underlying AMQP channel is not exposed by the `AmqpConsumer` and `AmqpPublisher`. I have deliberately not yet update the `node-dry-amqp` lib because there are some more issues around fault tolerance in the lib and with RabbitMQ in general, which require a complete refactor/rewrite of the  `node-dry-amqp` lib.

### DIFF
--- a/scheduler/src/api/amqp/datasourceConfigConsumer.ts
+++ b/scheduler/src/api/amqp/datasourceConfigConsumer.ts
@@ -1,5 +1,6 @@
 import * as AMQP from 'amqplib'
-import { AmqpConsumer } from '@jvalue/node-dry-amqp'
+import * as AmqpConnector from '@jvalue/node-dry-amqp/dist/amqpConnector'
+import { sleep } from '@jvalue/node-dry-basics'
 
 import Scheduler from '../../scheduling'
 import DatasourceConfig from '../datasource-config'
@@ -15,13 +16,27 @@ import {
 } from '../../env'
 
 export class DatasourceConfigConsumer {
-  private readonly consumer = new AmqpConsumer()
+  private connection?: AMQP.Connection
+  private channel?: AMQP.Channel
 
   constructor (private readonly scheduler: Scheduler) {
   }
 
   public async initialize (retries: number, backoff: number): Promise<void> {
-    await this.consumer.init(AMQP_URL, retries, backoff)
+    for (let i = 1; i <= retries; i++) {
+      try {
+        this.connection = await AmqpConnector.connect(AMQP_URL)
+        break
+      } catch (error) {
+        console.info(`Error initializing the AMQP Client (${i}/${retries}):
+        ${error}. Retrying in ${backoff}...`)
+      }
+      await sleep(backoff)
+    }
+
+    if (this.connection === undefined) {
+      throw new Error(`Could not connect to AMQP broker at ${AMQP_URL}`)
+    }
 
     const exchange = { name: AMQP_SCHEDULER_EXCHANGE, type: 'topic' }
     const exchangeOptions = {}
@@ -29,18 +44,30 @@ export class DatasourceConfigConsumer {
     const queueOptions = {
       exclusive: false
     }
-    await this.consumer.registerConsumer(exchange, exchangeOptions, queue, queueOptions, this.consumeEvent)
+
+    this.channel = await AmqpConnector.initChannel(this.connection, exchange, exchangeOptions)
+
+    await this.channel.assertQueue(queue.name, queueOptions)
+    await this.channel.bindQueue(queue.name, exchange.name, queue.routingKey)
+  }
+
+  public async startEventConsumption (): Promise<void> {
+    if (this.channel === undefined) {
+      throw new Error('Missing channel, AMQP client not initialized')
+    }
+
+    await this.channel.consume(AMQP_SCHEDULER_QUEUE, msg => {
+      this.consumeEvent(msg)
+        .catch(error => console.error(`Failed to handle ${msg?.fields.routingKey ?? 'null'} event`, error))
+    })
   }
 
   private readonly consumeEvent = async (msg: AMQP.ConsumeMessage | null): Promise<void> => {
     if (msg === null) {
       console.debug('Received empty event when listening on datasource config events - doing nothing')
-    } else {
-      await this.handleMsg(msg)
+      return
     }
-  }
 
-  private readonly handleMsg = async (msg: AMQP.ConsumeMessage): Promise<void> => {
     if (isUpdateOrCreate(msg)) {
       const event: DatasourceConfigEvent = JSON.parse(msg.content.toString())
       const datasource = event.datasource
--- a/scheduler/src/index.ts
+++ b/scheduler/src/index.ts
@@ -2,7 +2,7 @@ import type { Server } from 'http'
 import express from 'express'
 
 import Scheduler from './scheduling'
-import { initSchedulerWithRetry } from './initializer'
+import { setupInitialStateWithRetry } from './initializer'
 
 import {
   CONNECTION_RETRIES,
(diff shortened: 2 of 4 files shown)


lunedis (commented):
I looked through this PR and also your (very nice btw) fault tolerance tests, but cannot figure out how this changes how the scheduler can reach invalid state.

As far as I can see, you split up the initialization call into initalization and a seperate call to start listening, but you call them right after each other anyways (both times with await).

Can you explain how exactly this fixes the problem?

Thanks in advance!

sonallux:
Sure, two aspects are important to fix the issue:

### 1. The scheduler must start listen to datasource change events before requesting the initial state
This ensures that no changes are lost after the Adapter has create the initial state. The scheduler is must only create the RabbitMQ queue and setup the correct message routing to this queue, but it must not start consuming messages (because of 2.). This is done in the `DatasourceConfigConsumer.initialize()` method.

### 2. Datasource change events must only be applied after the initial state has been applied
This ensures that no change events are "overtaking" the initial state, which would lead an invalid state. The start of consuming change events is done by the new `DatasourceConfigConsumer.startEventConsumption()` method.

The only real change of my PR is that I have moved the start of consuming events after the initial state has been applied.

--- 

To make the steps more understandable I think it might be worth splitting the `DatasourceConfigConsumer.initialize()` method into two methods. One for creating the AMQP connection and the other for setting up the queue and the message routing. And I would also rename the `initSchedulerWithRetry()` function to `setupInitialState()` What do you think @lunedis ?

lunedis:
> Sure, two aspects are important to fix the issue:
> 
> ### 1. The scheduler must start listen to datasource change events before requesting the initial state
> This ensures that no changes are lost after the Adapter has create the initial state. The scheduler is must only create the RabbitMQ queue and setup the correct message routing to this queue, but it must not start consuming messages (because of 2.). This is done in the `DatasourceConfigConsumer.initialize()` method.
> 
> ### 2. Datasource change events must only be applied after the initial state has been applied
> This ensures that no change events are "overtaking" the initial state, which would lead an invalid state. The start of consuming change events is done by the new `DatasourceConfigConsumer.startEventConsumption()` method.
> 
> The only real change of my PR is that I have moved the start of consuming events after the initial state has been applied.

Ah, now I see it was well. Not sure how I missed it, but it's pretty clear how `startEventConsumption` is called after `initSchedulerWithRetry`.
Thanks for the explanation!

> To make the steps more understandable I think it might be worth splitting the `DatasourceConfigConsumer.initialize()` method into two methods. One for creating the AMQP connection and the other for setting up the queue and the message routing. And I would also rename the `initSchedulerWithRetry()` function to `setupInitialState()` What do you think @lunedis ?

I do not think splitting up the `initialize` method is necessary, I think grouping the creation of the AMQPConnector and the setup of the queues together is reasonable.

I do however agree that renaming `initScheudulerWithRetry` is a good idea, as I overlooked that function when reviewing your PR.
Kinda hard to guess what it does just from its name.

sonallux:
> I do not think splitting up the `initialize` method is necessary, I think grouping the creation of the AMQPConnector and the setup of the queues together is reasonable.
> 
> I do however agree that renaming `initScheudulerWithRetry` is a good idea, as I overlooked that function when reviewing your PR. Kinda hard to guess what it does just from its name.

Ok, I have already renamed `initScheudulerWithRetry` to `setupInitialStateWithRetry`.

lunedis (approved):
LGTM

georg-schwarz (approved):

PR #302, opened 2021-01-12
Adapter: refactor the configuration

sonallux:
- refactor the handling of configuration by using spring boot's `ConfigurationProperties` instead of `System.getenv(...)`
- remove useless annotations and code
- refactor all `System.{err|out}.println(...)` calls to use a real `Logger`
- add [lombok](https://projectlombok.org) to reduce boilerplate code


georg-schwarz (dismissed):
nice!

sonallux:
@georg-schwarz and @lunedis are you ok if I add [lombok](https://projectlombok.org) to the adapter? With lombok we can get rid of a lot of boilerplate code (getter, setter, constructors, ...) by just adding an annotation.

georg-schwarz:
Yeah, I'd appreciate less clutter in our java code base. What do you think @lunedis ?

lunedis:
Lombok looks interesting, no objections from me for using it.

sonallux:
I have added lombok and now we have ~500 lines less in the adapter code base 🚀

georg-schwarz:
Awesome! Much cleaner now!

georg-schwarz (approved):

PR #304, opened 2021-01-18
Adapter: update tools

sonallux:
- Update to JDK 15 and use the new text blocks in UnitTests for much better readability
- Update to gradle 6.7
- Update to JUnit 5
- Update adapter `Dockerfile` to use a docker image that already contains gradle, so the gradle wrapper must not download gradle everytime.


georg-schwarz (approved):
Awesome!

PR #306, opened 2021-01-19
Fix pulling out adapter version from properties file

georg-schwarz:


.github/workflows/ods.yml
```diff
@@ -302,7 +302,7 @@ jobs:
 
           IMAGE_ID=$(sed -n 's/^DOCKER_REGISTRY=//p' .env)/adapter
 
-          ADAPTER_VERSION=$(grep "version =" ./adapter/src/main/resources/application.properties | awk '{print $3}' | sed "s/[']//g")
+          ADAPTER_VERSION=$(grep "app.version=" ./adapter/src/main/resources/application.properties | sed "s/[=]/ /g" | awk '{print $2}' | sed "s/[']//g")
```
sonallux: I think you can shorten that to:

```suggestion
          ADAPTER_VERSION=$(grep "version" ./adapter/src/main/resources/application.properties | sed "s/^app.version=//g")
```

georg-schwarz: Good point, thanks!

lunedis (approved):
LGTM

PR #307, opened 2021-01-19
Add adapter-outboxer

sonallux:
Adds the outbox pattern to the adapter.
- Add outboxer service to adapter database
- Insert event into outbox table instead of publishing to RabbitMQ
- Remove all RabbitMQ stuff from the adapter
- Adjust integration and system test because event delivery does take slightly longer now


docker-compose.yml
```diff
@@ -55,6 +48,30 @@ services:
     environment:
       - POSTGRES_USER=adapterservice
       - POSTGRES_PASSWORD=admin
+    command:
+      - '-cwal_level=logical'
+      - '-cmax_wal_senders=1'
+      - '-cmax_replication_slots=1'
```
georg-schwarz, line 55: What's that about?

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/api/amqp/AmqpPublisher.java
```diff
@@ -39,21 +39,7 @@ public void publishImportFailure(Long id, String errMsg) {
...
-          Thread.sleep(properties.getAmqp().getPublishBackoff());
-        } catch (InterruptedException interruptedException) {
-          Thread.currentThread().interrupt();
-          throw new RuntimeException(interruptedException);
-        }
-        log.warn("Message publish failed ({}). Retrying in {}", retries, properties.getAmqp().getPublishBackoff());
-      }
-    }
-    log.error("Sending message {} to topic {} failed", message.toString(), topic);
+    var event = new OutboxEvent(topic, message);
```
georg-schwarz, line 42: I assume that the calling class will manage the transaction then? Just curious...

docker-compose.yml
sonallux: Debezium uses [Postgres's logical decoding feature](https://www.postgresql.org/docs/current/logicaldecoding-explanation.html) to get the write-ahead log content. This enables and configures that feature.

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/api/amqp/AmqpPublisher.java
sonallux: No, transactions are handled by the `@Transactional` annotations in the `DatasourceManager`. For example here: https://github.com/jvalue/open-data-service/blob/1e670cdf2d36aae12273a2a965e0dd3e84edfe21/adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/DatasourceManager.java#L49-L56

docker-compose.yml
georg-schwarz: please add a comment for that! ;-)

sonallux: Done!

georg-schwarz (dismissed):
LGTM - @lunedis pls also have a look ;-)

sonallux:
I have just reverted a change that I made because of testing purposes, The `adapter-db` is now no longer exposed by default.

lunedis (approved):
LGTM

georg-schwarz (approved):

PR #309, opened 2021-01-22
Datasource config preview

georg-schwarz:
Added a preview for import config. Not very elaborate, but should do the job. Required by Alex who will generate a schema in a further step based on a preview.

image


lunedis (approved):
LGTM

PR #311, opened 2021-01-27
Add pipeline-outboxer service

sonallux:
Adds the outbox pattern to the pipeline service.
- Add the outboxer service and connect it to the pipeline database
- Insert events into outbox table instead of publishing to RabbitMQ
- Remove all AMQP publishing logic

~~Edit: Somehow the tests are failing, but they pass locally for me...~~


pipeline/src/pipeline-config/pipelineDatabase.ts
```diff
@@ -0,0 +1,181 @@
...
+const GET_ALL_CONFIGS_STATEMENT = `
+  SELECT * FROM "${POSTGRES_SCHEMA}"."${CONFIG_TABLE_NAME}"`
+const GET_ALL_CONFIGS_BY_DATASOURCE_ID_STATEMENT = `
+  SELECT * FROM "${POSTGRES_SCHEMA}"."${CONFIG_TABLE_NAME}" WHERE "datasourceId" = $1`
+const DELETE_CONFIG_STATEMENT = `
+  DELETE FROM "${POSTGRES_SCHEMA}"."${CONFIG_TABLE_NAME}" WHERE "id" = $1 RETURNING *`
+const DELETE_ALL_CONFIGS_STATEMENT = `
+  DELETE FROM "${POSTGRES_SCHEMA}"."${CONFIG_TABLE_NAME}" RETURNING *`
+
+const CREATE_OUTBOX_TABLE_STATEMENT = `
```
georg-schwarz: Does it make sense to separate pipeline handling from pipelineEvent handling?

sonallux: I have thought of this, but did not do it because the initialization (e.g. creating the database tables) should be done in the same transaction in the [init function](https://github.com/jvalue/open-data-service/pull/311/files/9df84569714ae1e4ff32990fce656c53faf3242b#diff-7344c05a5acea5f112528cd03003ce272755d6f25349df9d2e898a57beaff7e1R78-R86).

sonallux: After thinking more about this, we could actually do this by splitting it into three parts: The init function, the pipeline config handling and the pipeline event handling. The init function would then just call a create table functions for the pipeline config and the pipeline event table

sonallux: I have just pushed a new commit containing the proposed split. It is now a much better separation of concerns 🚀

georg-schwarz (approved):

georg-schwarz (approved):

PR #314, opened 2021-01-28
Wp1

LechoDecho:
This feature should introduce an interface to later be used with the schema generation. 
As so there is now an additional step inside the datasource creation. This step should give the user the possibility to choose whether they want a generation and if so in which detail level they want it. 
On the other side, there is an API that will be used to get the generated schemas. Currently, there is no generation of a schema so the test JSON which is sent to the schema service is simply sent back.


schema/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+WORKDIR /build
+
+# Copy package*.json files first in order to make best use of docker layer caching
+COPY ./package*.json ./
+
+# npm clean slate install to get reproducible builds and quicker installs
+RUN npm ci
+
+# copy rest of the files
+COPY ./src ./src
```
georg-schwarz: pls copy src last since it changes most often (utilizes docker layer caching better)

schema/src/index.ts
```diff
@@ -0,0 +1,23 @@
+import express from 'express'
+var app = express()
+app.use(express.json())
+
+// define a route handler for the default home page
+app.post('/fastGen', (req: express.Request, res: express.Response) => {
```
georg-schwarz: pls rename routes to `/suggestion/fast` and `/suggestion/detailed` or something like that

schema/src/index.ts
```diff
@@ -0,0 +1,23 @@
...
+app.post('/detailedGen', (req: express.Request, res: express.Response) => {
+  console.log(req.body)
+  res.send(req.body)
+})
+
+app.get('/', (req: express.Request, res: express.Response) => {
+  res.send(200)
+})
+
+app.listen(8520, () => {
```
georg-schwarz: We should use standard port `8080` before merging and use the routing mechanism of traefik to reach the service.

ui/Dockerfile
```diff
@@ -29,6 +30,7 @@ RUN echo "VUE_APP_BASE_URL=$UI_BASE_URL" > .env \
     && echo "VUE_APP_STORAGE_SERVICE_URL=$STORAGE_SERVICE_URL" >> .env \
     && echo "VUE_APP_PIPELINE_SERVICE_URL=$PIPELINE_SERVICE_URL" >> .env \
     && echo "VUE_APP_NOTIFICATION_SERVICE_URL=$NOTIFICATION_SERVICE_URL" >> .env \
+    && echo "VUE_APP_SCHEMA_SERVICE_URL=$SCHEMA_SERVICE_URL" >> .env \
```
georg-schwarz, line 33: Nice! Good job digging in there and making it work! I know it is a little complicated...

ui/src/datasource/DatasourceCreate.vue
```diff
@@ -67,6 +67,10 @@ export default class DatasourceCreate extends Vue {
       type: 'JSON',
       parameters: {}
     },
+    dataSchema: {
```
georg-schwarz, line 71: I think we will eventually simplify to `dataSchema: <schema>`, that's the only thing the datasource config should need. Converting the data with a mode into the schema should happen independently on the ui, where it comes from shouldn't matter to the final config. Mode and the actual data will then only exist in the config step ui component

ui/src/datasource/datasourceRest.ts
```diff
@@ -13,6 +13,12 @@ const http = axios.create({
   transformResponse: []
 })
 
+const httpSchema = axios.create({
```
georg-schwarz, line 16: Pls extract the schema recommendation requests to another file like `schemaSuggestionRest.ts`

ui/src/datasource/edit/schema/DatasourceSchemaEdit.vue
```diff
@@ -0,0 +1,82 @@
...
+import Component from 'vue-class-component'
+import * as DatasourceREST from './../../datasourceRest'
+
+import { Emit, PropSync } from 'vue-property-decorator'
+import { requiredRule } from '../../../validators'
+import { DataSchema } from '../../datasource'
+
+@Component({ })
+export default class DatasourceSchemaEdit extends Vue {
+  private availableSchemaModes = ['NONE', 'FAST', 'DETAILED']
```
georg-schwarz: Please use this in the model `mode: "NONE" | "FAST" | "DETAILED"' to ensure these are the only viable values by typing

ui/src/datasource/edit/schema/DatasourceSchemaEdit.vue
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  private onGenerate (): void {
+    if (this.dataSchema.mode == 'NONE')
+      return
+    else if (this.dataSchema.mode == 'FAST') {
+      DatasourceREST.getSchemaFast(this.dataSchema)
+        .then((value) => {
+          this.dataSchema.data = value
+          console.log(value)
```
georg-schwarz: Please remove all console logs before converting from draft to actual PR (no problem during development)

georg-schwarz (commented):
Hi Alex,

looks really good until now! 

I think we will actually move the schema service into a different repository. That might make things a little more complicated. The link to this application will be done via the Docker URL, you can also point the build context in compose to the other local git repo then for easier development. I'll send you the repository link...

Please don't hesitate to ask for help if it doesn't work out of the box!

Cheers,
Georg

georg-schwarz:
`/schema` can be removed since migrated to another repository.

PR #315, opened 2021-02-03
Refactor StorageMQ service for more fault tolerance

### LINKED ISSUE #275: StorageMQ: Use node-dry-amqp
To have a consistent AMQP usage across all microservices, the StorageMQ service should also use the `node-dry-amqp` library.

sonallux:
- Add `node-dry-amqp`
- Update `node-dry-pg`
- ~~Stop creating the content table on pipeline created events. Instead, create the content table if needed when adding content.~~

Fixes #275


georg-schwarz:
I think we should still create the tables on pipeline events.. right now it might be doable as you are suggesting just in time, but as soon as we introduce data schemas that won't work any more since we will need some information of the pipeline config to create the tables.

sonallux:
> I think we should still create the tables on pipeline events.. right now it might be doable as you are suggesting just in time, but as soon as we introduce data schemas that won't work any more since we will need some information of the pipeline config to create the tables.

Ok, I was not aware of this. I am reverting that change.

georg-schwarz (approved):

PR #316, opened 2021-02-03
Scheduler: update node-dry-amqp

sonallux:


scheduler/src/api/amqp/datasourceConfigConsumer.ts
```diff
@@ -73,16 +48,13 @@ export class DatasourceConfigConsumer {
       const datasource = event.datasource
       datasource.trigger.firstExecution = new Date(event.datasource.trigger.firstExecution)
       this.scheduler.upsertJob(datasource)
-      return
```
georg-schwarz, line 76: When I compare to the old code at this point we have a different reading flow. Please try to always have the "happy path" clear visible. Example:
 ```java
xy;
if (xy == 1) {
  // unhappy path
  return;
}

if (xy == 2) {
  // unhappy path
  return;
}

// happy path
return xyz;
```

As you can see, the happy path us clearly visible by not being indented. Everything indented on the other hand is a unhappy path.

Your proposed code here is syntactically correct of course, but I'd prefer to be able to directly see the happy path ;-)

Comment (sonallux) [reply scheduler/src/api/amqp/datasourceConfigConsumer.ts:76]: Well, actually the code has multiple happy paths and a common action for all happy paths:
```ts
if (msg === null) {
  return // unhappy path
}

if (...) {
  // happy path
} else if (...) {
  // happy path
} else {
  // happy path
}
// common action for all three happy paths
return
```

With the old approach, the common action must be duplicated for each happy path. Because I wanted to avoid duplicating the common action. I changed the structure.

georg-schwarz: Ah stupid me, kind of looked very different for me on mobile view. Maybe not the best idea to do code reviews on a small screen xD 
I see your point now and agree!

georg-schwarz (approved):

PR #318, opened 2021-02-05
CI: prevent image publication step on forks

sonallux:
Currently, the image publication step is also executed on pushes to a fork's master branch. This will fail due to insufficient permissions (for example see [my fork](https://github.com/sonallux/open-data-service/actions?query=workflow%3A%22Open+Data+Service+%28ODS%29%22)). With this change, the image publication step is only executed on the `jvalue/open-data-service`  repository.

### DIFF
--- a/.github/workflows/ods.yml
+++ b/.github/workflows/ods.yml
@@ -301,7 +301,7 @@ jobs:
 
     needs: [systemtest]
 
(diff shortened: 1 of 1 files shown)


georg-schwarz (approved):
Awesome, thanks!

Issue #232, opened 2020-10-28
Adapter: ResponseStatusException message is not present in response body


sonallux:
The message of a `ResponseStatusException` is not added to the HTTP response body. I would expect the exception message to be present in the response body for easier debugging and error spotting.

#### Reproducible example
When adding a new Datasource (`PUT /api/adapter/datasources`) with an `id` already specified, the response does not contain the error message of the thrown `ResponseStatusException` from this line: https://github.com/jvalue/open-data-service/blob/8404c3761ac5f46cfbe3367253e2f61916eff3ea/adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/api/rest/v1/DatasourceEndpoint.java#L43 Instead the response body just contains this:
```json
{
    "timestamp": "2020-10-28T17:27:40.435+00:00",
    "status": 400,
    "error": "Bad Request",
    "message": "",
    "path": "/datasources"
}
```

sonallux:
Closed by #322

PR #322, opened 2021-02-15
adapter: DataImports API and refactoring

### LINKED ISSUE #135: Adapter: Get imports and imported data of datasource via API
Import does not contain data, but only id, timestamp, link to data (url)

API suggestion:
* /datasources/{id}/imports -> map of all imports (id, timestamp, link to data url)
* /datasources/{id}/imports/{id} -> import (id, timestamp, link to data url)
* /datasources/{id}/imports/latest -> latest data import
* /datasources/{id}/imports/{id}/data -> the real data

Implementation:
* managing imports and import data should be handled in sub-package of datasources (not adapters). Probably requires a fair bit of refactoring.
* /dataImport should return imported data directly, no caching of these kind of data imports

lunedis:
closes #135 

I have a very bad habit of not commiting while I work, so sorry in advance for this mess of a PR consisting of 1 huge commit.

This PR contains the following:

- Introduction of the concept of a "Data Import", which is the result of calling `/trigger` on a datasource. DataImports are saved in the database and can be accessed using new API endpoints (see README and related issue).
- Big exception related refactoring: The internal classes (`Adapter`, `DatasourceManager`, ...) throw only application-specific exceptions now. Those exceptions are translated into `RestResponseExceptions` with suitable status code in the endpoint classes. I got a bit carried away and thus this is done using a `Map` of exceptions to status codes and a fancy method that takes a custom lambda. Check the code review for a caveat.
- Some cleanup of imports related to using lombok for e.g. Equals and Hashcode methods.
- Updated readme, unit tests and integration tests for all of the above.


adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/api/rest/v1/DatasourceEndpoint.java
```diff
@@ -20,52 +22,48 @@
...
-  public void updateDatasource(
-    @PathVariable Long id,
-    @Valid @RequestBody Datasource updateConfig) {
-    try {
-      datasourceManager.updateDatasource(id, updateConfig);
-    } catch (IllegalArgumentException e) {
-      throw new ResponseStatusException(HttpStatus.NOT_FOUND, "Datasource needs to exist before updating", e);
-    }
+  public void updateDatasource(@PathVariable Long id, @Valid @RequestBody Datasource updateConfig) {
+    this.handleErrors(() -> { datasourceManager.updateDatasource(id, updateConfig); return null;});
```
lunedis: the `return null;` is necessary as the `handleErrors` function expects a `CheckedSupplier` which returns something.

It would be an option to overload that function expecting a (to be defined) `CheckedRunnable` that returns void, but that would mean more code duplication. It could have been so pretty :cry: ...

sonallux (commented):
Nice work introducing application specific exceptions. I really like that, it is much cleaner now!

Instead of using our own `handleErrors(...)` function, you could also have used these fancy Spring annotations: [`@ExceptionHandler`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/bind/annotation/ExceptionHandler.html) and [`@ControllerAdvice`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/bind/annotation/ControllerAdvice.html). [Here](https://www.baeldung.com/exception-handling-for-rest-with-spring) and [here](https://spring.io/blog/2013/11/01/exception-handling-in-spring-mvc) you can find some tutorials how to use these annotations if you want to go with the spring way of handling exceptions. This is just for reference, for me, it is also ok sticking with the  `handleErrors(...)` function.

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/api/rest/v1/Mappings.java
```diff
@@ -1,9 +1,22 @@
...
+import org.springframework.http.HttpStatus;
+
 public class Mappings {
   public static final String IMPORT_PATH = "/preview";
   public static final String RAW_IMPORT_PATH = "/preview/raw";
   public static final String FORMAT_PATH = "/formats";
   public static final String PROTOCOL_PATH = "/protocols";
   public static final String VERSION_PATH = "/version";
+
+  public static final Map<Class, HttpStatus> ERROR_MAPPING = Map.of(
```
georg-schwarz: I wonder if a status code annotation directly at the exception is better?
```java
@ResponseStatus(code = HttpStatus.BAD_REQUEST)
class CustomException extends RuntimeException {}
```
Probably you had some thoughts on that, do you mind sharing them? :)

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/model/exceptions/AdapterException.java
```diff
@@ -0,0 +1,10 @@
+package org.jvalue.ods.adapterservice.adapter.model.exceptions;
+
+import lombok.NoArgsConstructor;
+
+@NoArgsConstructor
```
georg-schwarz, line 5: I would have assumed that lombok has some help for exceptions, but seems like the PR is not yet merged: https://github.com/rzwitserloot/lombok/pull/2702

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/DatasourceManager.java
```diff
@@ -38,27 +41,25 @@ public Datasource createDatasource(Datasource config) {
...
   public Iterable<Datasource> getAllDatasources() {
     return datasourceRepository.findAll();
   }
 
   @Transactional
-  public void updateDatasource(Long id, Datasource update) throws IllegalArgumentException {
-    Datasource existing = datasourceRepository.findById(id)
-      .orElseThrow(() -> new IllegalArgumentException("Datasource with id " + id + " not found"));
+  public void updateDatasource(Long id, Datasource update) throws DatasourceNotFoundException {
+    Datasource existing = datasourceRepository.findById(id).orElseThrow(() -> new DatasourceNotFoundException(id));
```
georg-schwarz, line 54: I think these chains are hard to read.. I guess we should introduce some linter or so in the future xD

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/api/rest/v1/DatasourceEndpoint.java
```diff
@@ -20,52 +22,48 @@
 public class DatasourceEndpoint {
   private final DatasourceManager datasourceManager;
 
+  private <T> T handleErrors(CheckedSupplier<T> function) throws ResponseStatusException {
```
georg-schwarz: duplicate with other endpoints => should we use a convert method and place it next to the mapping?

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/model/DataImport.java
```diff
@@ -0,0 +1,61 @@
...
+
+  @Id
+  @GeneratedValue
+  private Long id;
+
+  private byte[] data;
+  
+  private Date timestamp;
+
+  @ManyToOne(fetch = FetchType.EAGER)
```
georg-schwarz, line 26: That means the datasource is always also retrieved, right? In general, not a bad idea, but why do we not serialize it then? Just a little confused, do we need it?

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/model/Datasource.java
```diff
@@ -39,6 +40,11 @@
   @NotNull
   private DatasourceTrigger trigger;
 
+  @OneToMany(cascade = CascadeType.ALL, mappedBy = "datasource")
+  @JsonIgnore
+  @EqualsAndHashCode.Exclude // needed to avoid an endless loop because of a circular reference
+  private Set<DataImport> dataImports;
```
georg-schwarz, line 46: When are they fetched? Always on a get? Might lead to a huge overload once we had run 100 imports for a source... Maybe we should not link these concepts at all and just provide a datasourceId in the DataImport?

adapter/src/main/java/org/jvalue/ods/adapterservice/adapter/api/rest/v1/Mappings.java
lunedis: > I wonder if a status code annotation directly at the exception is better?
> 
> Probably you had some thoughts on that, do you mind sharing them? :)

I honestly had no clue about the way you are "supposed" to do it in Spring. Gonna have to take a look at it in detail. The solution I went for was the first one that came to mind to avoid repetition in every endpoint method.

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/DatasourceManager.java
lunedis: I actually kinda like chains like that, guess that comes from my preference for functional programming :grin:.

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/api/rest/v1/DatasourceEndpoint.java
lunedis: There are two different Mappings classes though, one for the Adapter, and one for the Datasources.

I have not found a good way to avoid this duplication yet. Someone got any ideas?

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/model/DataImport.java
lunedis: We need its `id` to create the URI to the data in `DataImportMetadata`.

Also the annotations are necessary for Spring to create the database tables and fields properly I think.

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/model/Datasource.java
lunedis: I put in a `fetch = FetchType.LAZY` which should resolve your worries about performance.

> Maybe we should not link these concepts at all and just provide a datasourceId in the DataImport?

Usually I prefer things doing manually as well, I am not sure if all the Spring SQL magic still works then though. And to be honest, the foreign keys generated by using those annotations are kinda neat.

sonallux: If we do not want this field, we can safely remove it, because the `datasource` field in the `DataImport` class is still annotated with `@ManyToOne`. This should be sufficient for Hibernate the generate the foreign keys.

lunedis: I use that field to generate the result of the `/datasources/{id}/imports` call though. But of course that could also be done using a function in the repository.

sonallux: Yeah, I would definitely go with a separate `findAllByDatasourceId` function in the DataImportRespository. This will be much faster because querying all other properties of the Datasource is useless when just return the list of DataImports.

lunedis: Problem I see with this approach: 

We want to differentiate between "return empty array because no data imports were found for the data source" and "return 404 because datasource does not exist".

If we use a function called `findAllByDatasourceId` it would just return nothing if the datasource does not exist at all.
So we kinda need to fetch the datasource details anyways (or check if it exists, which should not be a big difference).

sonallux: Ah, I haven't considered that 🙈

georg-schwarz: Good point.. I am kind of unsure what is best...

I guess lazy loading is misleading when reading code - or at least I'd assume the relation is fetched since I don't work with Spring too often. So personally, I'd not have the reference and make a separate call to prevent these uncertainties from happening. The readability would be clear: 

* if datasource does not exist => return 404
* if imports do not exist => return []

The obvious disadvantage: two database requests instead of one..

sonallux: > The obvious disadvantage: two database requests instead of one..

Well, I think having `@OneToMany` with `FetchType.LAZY` will also be two database requests...

lunedis:
I updated the PR and changed the exception handling to be more spring-y (?, springful? springesque?).

@sonallux could you please take a look?

Also that was a very short endeavor into the functional side of Java... :cry:

georg-schwarz (dismissed):

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/api/rest/v1/DataImportEndpoint.java
```diff
@@ -0,0 +1,64 @@
...
+public class DataImportEndpoint {
+  private final DatasourceManager datasourceManager;
+
+  @GetMapping("/{datasourceId}" + Mappings.DATAIMPORT_PATH)
+  @Transactional
+  public Iterable<DataImport.MetaData> getDataImports(@PathVariable Long datasourceId)
+      throws DatasourceNotFoundException {
+    Collection<DataImport> dataImports = datasourceManager.getDataImportsForDatasource(datasourceId);
+    Collection<DataImport.MetaData> metadatas = dataImports.stream()
+        .map((DataImport dataImport) -> dataImport.getMetaData()).collect(Collectors.toList());
```
sonallux:
```suggestion
        .map(DataImport::getMetaData).collect(Collectors.toList());
```

adapter/src/main/java/org/jvalue/ods/adapterservice/datasource/api/rest/v1/DataImportEndpoint.java
```diff
@@ -0,0 +1,64 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+import lombok.AllArgsConstructor;
+
+@AllArgsConstructor
+@RestController
+@RequestMapping(Mappings.DATASOURCE_PATH)
```
sonallux:
```suggestion
@RequestMapping(Mappings.DATASOURCE_PATH + "/{datasourceId}" + Mappings.DATAIMPORT_PATH)
```
I think we can specify more of the common path prefix here to make the `@GetMapping(...)` shorter and remove duplication.

sonallux (commented):
Just two very minor things, otherwise LGTM. The exception handling is now much more springish 👍 

This PR also fixes #232 🚀

sonallux (approved):

georg-schwarz (approved):

georg-schwarz:
Nice! I think this feature is a good step forward :)

PR #323, opened 2021-02-16
Schema Recommendation UI component (empty)

LechoDecho:
* Connection to schema recommendation backend
* Only show UI component if schema service is reachable, otherwise hide.


ui/src/datasource/DatasourceForm.vue
```diff
@@ -52,11 +52,12 @@
         :complete="dialogStep > 3"
         step="3"
       >
```
georg-schwarz, line 56: Can we add a `v-if` that only shows this step if the schema service is reachable?

ui/src/datasource/edit/schema/DatasourceSchemaEdit.vue
```diff
@@ -0,0 +1,80 @@
...
+    this.emitValue()
+    this.emitValid()
+  }
+
+  private onGenerate (): void {
+    if (this.dataSource.mode == Mode.NONE)
+      return
+    else if (this.dataSource.mode == Mode.FAST) {
+      SchemaSuggestionREST.getSchemaFast(this.dataSource.dataSchema)
+        .then((value) => {
```
georg-schwarz: pls use async/await syntax to be consistent with the existing code

georg-schwarz (changes requested):
Hi Alex, 

I left two comments.
Please also sign the CLA with our CLA-Assistant...

Cheers,
Georg

ui/src/datasource/DatasourceForm.vue
```diff
@@ -47,16 +50,36 @@
           @stepChanged="dialogStep = $event"
         />
       </v-stepper-content>
+      <v-stepper-content 
+        v-if="!updateIsSchemaAlive"
```
georg-schwarz: pls use `isSchemaAlive` everywhere in HTML instead of `updateIsSchemaAlive`

ui/src/datasource/DatasourceForm.vue
```diff
@@ -93,39 +134,57 @@
...
   private datasource: Datasource | undefined
 
+  private isSchemaAlive: boolean = false
   private required = requiredRule
 
   get isLoading (): boolean {
     return this.datasource === undefined
   }
 
+  get updateIsSchemaAlive (): boolean {
```
georg-schwarz: Changes the state (`isSchemaAlive`), so should not be a getter. Please make it into a normal async private function that uses the `async/await` syntax. Use the Vue lifecycle hooks, like `mounted` to call the update method when the component gets rendered.

ui/src/datasource/DatasourceForm.vue
```diff
@@ -47,16 +50,36 @@
...
+        step="2"
+      >
+        <adapter-config
+          v-model="datasource"
+          @validityChanged="validStep2 = $event+1"
+        />
+        <stepper-button-group
+          :step="2"
+          :next-enabled="validStep2"
+          @stepChanged="dialogStep = $event+1"
```
georg-schwarz: When I saw it right, the only difference here is that you add `+1` to skip the next step, right? We should be able to remove the duplicate by using an expression like `isSchemaAlive? $event : $event + 1`

georg-schwarz (approved):

PR #326, opened 2021-02-26
Update dependencies

georg-schwarz:


ui/package.json
```diff
@@ -9,42 +9,42 @@
...
     "eslint-plugin-node": "^11.1.0",
-    "eslint-plugin-promise": "^4.2.1",
-    "eslint-plugin-vue": "^7.0.1",
-    "monaco-editor-webpack-plugin": "^1.9.0",
-    "sass": "^1.26.8",
-    "ts-jest": "^24.3.0",
-    "typescript": "^3.9.7",
+    "eslint-plugin-promise": "^4.3.1",
+    "eslint-plugin-vue": "^7.6.0",
+    "monaco-editor-webpack-plugin": "^3.0.0",
```
sonallux: We can not update `monaco-editor-webpack-plugin` to `3.0.0` because it only works with `monaco-editor` `0.22.x`. We are using `vue-monaco` which currently is using `monaco-editor` `0.20.0`. Therefore we must stick with version `1.9.0`. Also see the [version compatibility matrix](https://github.com/microsoft/monaco-editor-webpack-plugin#version-matrix)  for `monaco-editor-webpack-plugin`.

sonallux (changes requested):

ui/package.json
georg-schwarz: I think I found a way how it should work in commit 2988db7

georg-schwarz: Not sure though, if it is a good idea to use this workaround..

sonallux: Well, even if the build now works, I do not think it is a good idea to fix this that way.

With your change we will have both version of `monaco-editor` in our dependency tree and `vue-monaco` will still use the old version `0.20.0`...

sonallux: I also think it is not very nice, because with your change we will have both versions of `monaco-editor` in the dependency tree. The older version is referenced by `vue-monaco` and the newer by `monaco-editor-webpack-plugin`. 

I just had a look into `vue-monaco` and I think your workaround will indeed work because `monaco-editor-webpack-plugin` will replace the old `monaco-editor` version from `vue-monaco` during build time by the new version.

Your change will also result in a longer installation and build times, because two `monaco-editor` versions must be downloaded and `monaco-editor` is a very large package.

---

`vue-monaco` is actually just a tiny wrapper around the `monaco-editor` with only 165 loc. To solve this issue we could also just copy the files from `vue-monaco` and remove the dependency. `vue-monaco` is MIT licenced so this is not an issue, is it?

georg-schwarz: Hm, I agree that the solution is not optimal... I'd not copy it because we would not profit from new versions.. Well, I think we don't require the version upgrade by all means, so I'll downgrade `monaco-editor-webpack-plugin` back to `1.9.0`

lunedis (approved):
LGTM

PR #344, opened 2021-06-30
Refactor system test: nodejs excecution

georg-schwarz:
Refactor system tests to directly run on nodejs and not docker. This leads to the same setup as local development execution. The tests starts all required containers itself.


system-test/jest.config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "ODS SystemTests",
+  "globalSetup": "<rootDir>/src/setup.js",
+  "globalTeardown": "<rootDir>/src/teardown.js"
```
sonallux, line 4: I really like this approach using jest's `globalSetup` and `globalTeardown` hooks 🚀

system-test/src/util/env.js
```diff
@@ -0,0 +1,41 @@
+const DOCKER_COMPOSE_FILE = '../docker-compose.yml -f ../docker-compose.st.yml'
```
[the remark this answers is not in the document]
georg-schwarz: Thx!

system-test/src/setup.js
```diff
@@ -0,0 +1,24 @@
...
+const { DockerCompose } = require('./util/docker-compose')
+const { waitForServicesToBeReady } = require('./util/waitForServices')
+
+async function setupTestEnv () {
+  const dockerCompose = DockerCompose(DOCKER_COMPOSE_FILES, DOCKER_ENV_FILE)
+  const docker = new Docker({})
+
+  if ((await docker.listContainers()).length !== 0) {
+    throw Error('Can not execute restart test if other containers are running')
+  }
```
lunedis, line 13: Is it worth it to pull in the `dockerode` dependency just for this single call?

system-test/src/util/docker-compose.js
```diff
@@ -0,0 +1,33 @@
...
+  const escapedTestName = expect.getState().currentTestName.split(/,| |:/).join('_')
+  const logFile = `${LOGS_DIRECTORY}/${escapedTestName}.log`
+  if (!fs.existsSync(LOGS_DIRECTORY)) {
+    fs.mkdirSync(LOGS_DIRECTORY)
+  }
+  const fd = fs.openSync(logFile, 'w')
+  for (const service of services) {
+    const logs = await dockerComposeFn(`logs --no-color ${service}`)
+    fs.writeSync(fd, logs.stdout)
+  }
```
lunedis, line 26: Very minor nitpick, and very likely does not matter at all performance wise, but there must be a way to just pipe that through without saving the whole log in a variable, right? :smile:

lunedis (approved):
Looks very nice, although honestly I did not fully understand the reasoning to change the system test process. Were there startup problems?

This system test would now be the only job in the whole CI pipeline that runs directly on node, and not just executes docker (compose) calls.

georg-schwarz:
> Looks very nice, although honestly I did not fully understand the reasoning to change the system test process. Were there startup problems?
> 
> This system test would now be the only job in the whole CI pipeline that runs directly on node, and not just executes docker (compose) calls.

The reason for the change is development experience. You can now debug the system tests quite easily since the test setup is done by jest (just use the Jest extension for VSCode and you can click on the test you want to debug).

Additionally, the timing of the startup of the components should be improved since explicitly done after each other. We always had some timing issues in CI, so I hope this is improved as well with this change.

system-test/src/util/docker-compose.js
georg-schwarz: It should be possible in general but `docker-compose logs` already returns a string and not a stream, so I don't think it makes any difference here.

lunedis:
> > Looks very nice, although honestly I did not fully understand the reasoning to change the system test process. Were there startup problems?
> > This system test would now be the only job in the whole CI pipeline that runs directly on node, and not just executes docker (compose) calls.
> 
> The reason for the change is development experience. You can now debug the system tests quite easily since the test setup is done by jest (just use the Jest extension for VSCode and you can click on the test you want to debug).
> 
> Additionally, the timing of the startup of the components should be improved since explicitly done after each other. We always had some timing issues in CI, so I hope this is improved as well with this change.

I see. All good points, so LGTM :+1:

PR #346, opened 2021-07-07
System-test: replace Dockerode with custom code

georg-schwarz:
Follow-up on #344 

Remove dockerode dependency and use exec to show if containers are still running.


lunedis (approved):
LGTM :rocket:

Issue #347, opened 2021-08-09
RabbitMQ 3.9 requires configuration using file, not environment variables


lunedis:
We are currently using the RabbitMQ-image in version `rabbitmq:3-management-alppine`.

Starting from version 3.9, RabbitMQ no longer supports being configured using environment variables.

See this rabbitmq docker issue: https://github.com/docker-library/rabbitmq/issues/508

georg-schwarz:
Suggested hotfix: specify version `rabbitmq:3.8.19-management-alpine`
Update to `latest` again after we fixed the issue and provided a conf file.

PR #348, opened 2021-08-09
hotfix: set rabbitmq version to 3.8

lunedis:
hotfixes #347 .

We need a better solution though of course.


lunedis:
Ahh ok, I found the service we need the `management`-version for :smile:

georg-schwarz (approved):
Thx for the debugging! :)

PR #352, opened 2021-08-28
Introduce consumer-driven contract testing using Pact

felix-oq:
This PR introduces consumer-driven contract testing (CDCT) to the project. It is implemented using [Pact](https://docs.pact.io/).
Up to now I managed to implement a test for a single RESTful interaction where the UI is the consumer and the pipeline-service is the provider.
For now, the CI integration of CDCT is planned to be realized without the help of [Pact Broker](https://docs.pact.io/pact_broker/docker_images). The Pact Broker will probably be added later in order to compare both approaches.


georg-schwarz (approved):
Nice!

PR #355, opened 2021-09-10
Notification: Disable ssl for database connection

knusperkrone:
Necessary for the K8 database cluster, as SSL needs to be disabled


georg-schwarz (dismissed):

knusperkrone:
@georg-schwarz  ping, 

Forgot the pipeline-service

georg-schwarz:
Currently one of the fault tolerance tests are failing (also on main). I think we can merge nevertheless in this case since they should not be affected by this PR, right?

georg-schwarz (approved):

knusperkrone:
At least, I don't see any interference with  the fault tolerance test and this MR. :)

PR #357, opened 2021-09-17
Allow self signed certs

knusperkrone:
Changed to DB operator again. It uses a self-signed certificate, I disabled it for the previous microservices


georg-schwarz (dismissed):

georg-schwarz:
Integration tests are failing due to changes: `Failed to start notification service: Error: The server does not support SSL connections`

PR #361, opened 2021-10-01
Feature/351 upgrade typescript+eslint

f3l1x98:
**For following projects:**
- notification
- pipeline
- scheduler
- storage/storage-mq
- ui

**Done:**
- Upgraded typescript to 4.4.2
- Switched eslint to jvalue/eslint-config
- Fixed linter issues caused by new/stricter linter config
- Updated vscode settings file to use prettier on save
- Extracted fix in linter command to separate new command
- Added vetur config for vetur support in case project was opened in root level
- Added VSCode localhistory extension files to gitignore


georg-schwarz (approved):
Awesome! Thanks for the hard and diligent work ;)

Issue #360, opened 2021-10-01
CDC Test and Pact Readme


georg-schwarz:
As a contributor I'd like to have a Readme where I can learn about the cdc setup with pact in order to have a fast introduction into the topic.

Is should contain:
• What are cdc tests
• How to run them
• Where to find them in the project
• Steps to update and create them

@felix-oq

felix-oq:
@georg-schwarz Please take a look at this draft: https://github.com/felix-oq/open-data-service/blob/cdct-improve-workflows/doc/consumer-driven-contract-testing/README.md

Does it cover the topics detailed enough? Is the location of the Readme within the project fine?

georg-schwarz:
Looks good! Location is also fine.

Maybe add a little section with comparison to integration tests? Or what they test and what they don't test?

felix-oq:
@georg-schwarz By now, the issue should be solved by the PR #368

PR #362, opened 2021-10-01
Implement message-based CDC-Test between Notification and Pipeline

felix-oq:
Introduces the message-based CDC-Test between Notification and Pipeline. For the pipeline service, a rather complex test setup is required because the amqp messages are published by the outboxer service of the Pipeline. Because of that, the provider-side test of the Pipeline requires its database, its outboxer service and a rabbitmq instance to be present.


felix-oq:
For some reason, old commits from previous pull requests show up aswell. This probably occurs due to my feature branch workflow in the forked repository.

georg-schwarz:
Pls rebase on main, there were a few changes in the recent minutes.

Indeed, the setup is a little more elaborate than before. What do you think about the test setup is handled with docker-compose? I know we use it quite often in the project, but to me it feels quite clunky and we already started refactoring it. In the system-test we handled it like that: use a test setup script for starting all the services instead of using the docker-compose commands manually... So you can just start the whole test with `npm run test`.

https://github.com/jvalue/ods/blob/main/system-test/src/setup.js

What do you think? Should that be applied here as well?

felix-oq:
> Pls rebase on main, there were a few changes in the recent minutes.

Done 
> Indeed, the setup is a little more elaborate than before. What do you think about the test setup is handled with docker-compose? I know we use it quite often in the project, but to me it feels quite clunky and we already started refactoring it. In the system-test we handled it like that: use a test setup script for starting all the services instead of using the docker-compose commands manually... So you can just start the whole test with `npm run test`.
> 
> https://github.com/jvalue/ods/blob/main/system-test/src/setup.js
> 
> What do you think? Should that be applied here as well?

Such complex test setups should not be common regarding CDC-Tests, but there are cases where it is not avoidable (such as the outboxer services of Pipeline and Adapter and the PostgREST service of Storage). Using test setups written in JavaScript might be a bit overkill, because in order to run the CDC-Tests you just need to execute some docker-compose commands and wait until the execution finishes. For the System Test the mightiness of JavaScript comes in handy, for example to additionally send requests that are required within the test scenarios. For CDC-Testing a simple script file (e.g. bash) would be completely sufficient to run the tests in my opinion.

The test setup will be further improved in an upcoming PR. In that PR I could also add some scripts for an easier execution of the CDC-Tests and the desired documentation (issue #360).

georg-schwarz (approved):

PR #369, opened 2021-10-08
Bugfix/353 fe be pipeline incompatibility

f3l1x98:
- Fixed the REST endpoint to accept pipeline array
- Refactored the action to update the specific elements of the pipelines state variable
- Added vuex getter for retrieving all pipelines of given datasourceId from state
- Adjusted + enabled pact test


ui/src/pipeline/pipeline.module.ts
```diff
@@ -18,12 +23,33 @@ export default class PipelineModule extends VuexModule {
     PIPELINE_SERVICE_URL,
   );
 
+  get getPipelinesByDatasourceId(): (datasourceId: number) => Pipeline[] {
```
georg-schwarz: Starting the name of a get method with "get" is ambiguous. You use a getter method like a prop `x.getPipelinesByDatasourceId' without braces in the end. Either make it a normal function starting with "get", or name it like a property.

f3l1x98: Done

ui/src/pipeline/pipeline.module.ts
```diff
@@ -18,12 +23,33 @@ export default class PipelineModule extends VuexModule {
     PIPELINE_SERVICE_URL,
   );
 
+  get pipelinesByDatasourceId(): (datasourceId: number) => Pipeline[] {
+    return (datasourceId: number): Pipeline[] => {
```
georg-schwarz: ah sorry, my bad. Since it has a parameter it has to be a regular function "getPipelinesByDatasourceId(..)" instead of a getter.

f3l1x98: I wrote it as a gettter in order for Vuex to identify it as a getter and as far as I know this is the only way to write vuex getters with function params (by returning a function with the paramter).
So no it does not have a parameter, but the returned function does.

If there is a better way to write these type of vuex state getters I will change it accordingly.

lunedis:
@felix-oq Tests wurden wieder einkommentiert.

ui/src/pipeline/pipeline.consumer.pact.fixtures.ts
```diff
@@ -20,6 +20,19 @@ export const examplePipeline: Pipeline = {
...
+  metadata: {
+    author: 'some other author',
+    description: 'some other description',
+    displayName: 'some other display name',
+    license: 'some other license',
+  },
+  transformation: {
+    func: 'some other function',
+  },
+};
```
felix-oq: An additional example pipeline is not necessary. When writing consumer-side tests, the actual data doesn't matter, it's all about the structure of the requests and responses. During provider verification for example, usually just the types of the attributes in the response are checked. Their actual value does not matter.
```suggestion
```

ui/src/pipeline/pipeline.consumer.pact.fixtures.ts
```diff
@@ -77,13 +90,35 @@ export function getByDatasourceIdRequest(datasourceId: number): RequestOptions {
...
+  headers: {
+    'Content-Type': 'application/json; charset=utf-8',
+  },
+  body: [
+    like(examplePipeline),
+    like(examplePipeline2),
+    like(Object.assign({}, examplePipeline, { id: 3 })),
+  ],
+};
+
```
felix-oq: The additional response is not required for the test. Take a look at some of the following comments for an explanation.
```suggestion
```

ui/src/pipeline/pipeline.consumer.pact.test.ts
```diff
@@ -179,31 +182,36 @@ pactWith(options, provider => {
...
+
+          expect(pipelines).toStrictEqual([]);
         });
       });
 
       describe('with NaN as requested datasource id', () => {
         beforeEach(async () => {
           await provider.addInteraction({
-            state: 'any state',
+            state: 'pipelines with datasource id NaN return all pipelines',
```
felix-oq: The state of the provider can be arbitrary for this tested interaction, i.e. it does not matter whether any pipelines exist or not on the provider side.
When sending `NaN` as datasource id, the response should always be an empty array because, as far as I know, there can't be pipelines with a datasource id of `NaN`.

```suggestion
            state: 'any state',
```

ui/src/pipeline/pipeline.consumer.pact.test.ts
```diff
@@ -179,31 +182,36 @@ pactWith(options, provider => {
...
             uponReceiving: getByDatasourceIdRequestTitle(NaN),
             withRequest: getByDatasourceIdRequest(NaN),
-            willRespondWith: badRequestResponse,
+            willRespondWith: getByDatasourceIdGetAllResponse,
```
felix-oq: According to the previous comment, the response should contain an empty array in this interaction:

```suggestion
            willRespondWith: getByDatasourceIdEmptyResponse,
```

ui/src/pipeline/pipeline.consumer.pact.test.ts
```diff
@@ -179,31 +182,36 @@ pactWith(options, provider => {
...
-          ).rejects.toThrow(Error);
+        it('returns pipeline array containing all pipelines', async () => {
+          const pipelines = await restService.getPipelineByDatasourceId(NaN);
+
+          expect(pipelines).toHaveLength(3);
+          expect(pipelines[0]).toStrictEqual(examplePipeline);
+          expect(pipelines[1]).toStrictEqual(examplePipeline2);
+          expect(pipelines[2]).toStrictEqual(
+            Object.assign({}, examplePipeline, { id: 3 }),
+          );
```
felix-oq: Again as previously commented, the resulting array should be empty:

```suggestion
          expect(pipelines).toStrictEqual([]);
```

pipeline/src/api/rest/ui.provider.pact.test.ts
```diff
@@ -117,6 +117,8 @@ describe('Pact Provider Verification', () => {
         'pipeline with id 1 does not exist': setupEmptyState,
         'pipelines with datasource id 2 exist': setupSomePipelineConfigs,
         'pipelines with datasource id 2 do not exist': setupEmptyState,
+        'pipelines with datasource id NaN return all pipelines':
+          setupSomePipelineConfigs,
```
felix-oq: The additional provider state is not necessary. An explanation can be found in the following comments.

```suggestion
```

Also as a little hint, reagrding titles of provider states, the names should just imply in which state the provider should be for provider-side test execution. They shouldn't make any assumptions on the behaviour of the provider service.

ui/src/pipeline/pipeline.consumer.pact.test.ts
```diff
@@ -150,8 +153,7 @@ pactWith(options, provider => {
       });
     });
 
-    // TODO do not skip these tests anymore as soon as issue #353 is solved
-    describe.skip('getting a pipeline by datasource id', () => {
+    describe('getting pipelines by datasource id', () => {
       describe('when a pipeline with the requested datasource id exist', () => {
```
felix-oq: I noticed a typo here, I guess that was my bad:
```suggestion
      describe('when a pipeline with the requested datasource id exists', () => {
```

ui/src/pipeline/pipeline.consumer.pact.test.ts
```diff
@@ -164,10 +166,11 @@ pactWith(options, provider => {
...
         });
 
-        it('returns the requested pipeline', async () => {
-          const pipeline = await restService.getPipelineByDatasourceId(id);
+        it('returns the requested pipelines', async () => {
+          const pipelines = await restService.getPipelineByDatasourceId(id);
 
-          expect(pipeline).toStrictEqual(examplePipeline);
+          expect(pipelines).toHaveLength(1);
+          expect(pipelines[0]).toStrictEqual(examplePipeline);
```
felix-oq: Can be simplified:
```suggestion
          expect(pipelines).toStrictEqual([examplePipeline]);
```

felix-oq (changes requested):
I found some mistakes in the CDC-Tests. I hope the comments help to understand and fix the problems.

ui/src/pipeline/pipeline.consumer.pact.fixtures.ts
f3l1x98: But if I am not mistaken the next step should check their value:
https://github.com/jvalue/ods/blob/acda73c1fb05c7f5204463d9dc150dd1d4b93975/ui/src/pipeline/pipeline.consumer.pact.test.ts#L210-L211

ui/src/pipeline/pipeline.consumer.pact.test.ts
f3l1x98: You are right that there will be no pipelines with datasourceId `NaN`, but the backend api returns all pipelines, in case the datasourceId param is `NaN`
https://github.com/jvalue/ods/blob/acda73c1fb05c7f5204463d9dc150dd1d4b93975/pipeline/src/api/rest/pipelineConfigEndpoint.ts#L101-L118

ui/src/pipeline/pipeline.consumer.pact.fixtures.ts
felix-oq: Yes that's right, but there the interpreted response values (in this case objects in `pipelines`) are compared with the predefined response values (here `examplePipeline` and `examplePipeline2`). If the UI interprets the response correctly, they should always match, because the attribute values of the interpreted response and the predefined response are always the same. Thus the actual attribute values can be arbitrarily chosen.
This also means, that blowing up the response array by additional elements does not add any value to the test.

ui/src/pipeline/pipeline.consumer.pact.test.ts
felix-oq: Oh you are right, I missed that. Intuitively I would have expected an empty array or a bad request status as response in that case. @georg-schwarz is there a reason for that behavior?

ui/src/pipeline/pipeline.consumer.pact.fixtures.ts
f3l1x98: Ah, ok thanks

ui/src/pipeline/pipeline.consumer.pact.test.ts
georg-schwarz: I agree. Either `[]` or 400 makes sense.

ui/src/pipeline/pipeline.module.ts
georg-schwarz: Hm, probably it should not be part if the state at all then? Not sure though, I'm not too deep into state management in UI. @MBuchalik do you have an opinion?

MBuchalik: I am not that familiar with Vuex, so I have a few questions here and there...

First of all: Am I missing something or are you never calling your getter anywhere? Is it common practice to provide a pair of getters and setters in Vuex? If so, then your approach looks ok to me. At least at the first glance, I wouldn't know another approach for this. The only thing that I would probably do is to provide a small explanatory comment above the getter. (But that is up to you.)

If it is not necessary to provide a pair of getters and setters, then you could:
- Just leave it out.
- Provide a getter that returns all pipelines. Then you could either filter them directly in your component. Or you could export a helper function somewhere that does this job.

f3l1x98: Ok thanks.
I will remove it, because I think that was something for the future (I am not 100% myself anymore).
If we need it at some point we can just add it again (or the other suggestion)

ui/src/pipeline/pipeline.consumer.pact.test.ts
f3l1x98: @georg-schwarz I would do this change in this branch together with the resulting pact changes

f3l1x98: @felix-oq The changes are pushed and due to the API change I have applied all of your pact changes

georg-schwarz (approved):
Any need to update the PR according to the recent changes on main?

f3l1x98:
> Any need to update the PR according to the recent changes on main?

I don't think so.

PR #370, opened 2021-10-08
created rabbitmq.conf

f3l1x98:
- moved username and password into config file
- mounted config file in docker compose
- switched back to rabbitmq:3-management-alpine

See #347


georg-schwarz (approved):
Awesome, thx!!

PR #373, opened 2021-10-18
Implement CDCT for all interactions between UI and Storage

felix-oq:


ui/src/storage/storageRest.ts
```diff
@@ -1,34 +1,40 @@
...
+  async getStoredItem(
+    pipelineId: number,
+    storageItemId: number,
+  ): Promise<StorageItem | null> {
+    const response = await this.http.get(
+      `/${pipelineId}?id=eq.${storageItemId}`,
+    );
+    const responseData = response.data as StorageItem[];
+    return responseData.length === 0 ? null : responseData[0];
+  }
```
felix-oq: To the reviewer(s): please take a look at this change in particular.

Previously, the response data was also interpreted as an array, but the element at index 0 was always returned, regardless if the array was empty or not. This resulted in the resolved return value possibly being `undefined` which is not obvious given the return type `Promise<StorageItem>` of the function.

I came up with two possible approaches to fix the issue:

- As I've done it here: Changing the return type to `Promise<StorageItem | null>`, checking the length of the array and returning null if the array is empty
- Checking the length of the array and throwing an error if the array is empty

The function `getStoredItem` is only used here:
https://github.com/felix-oq/ods/blob/10bc74919264fd787a3021ed0bd24832940ef207/ui/src/storage/StorageItemView.vue#L61-L66

Can we keep the current fix or do you see any problems?

georg-schwarz: Nice, good catch!

I think following our pattern we used in other places, we should do `Promise<StorageItem | undefined>` unstead of `null`. We should avoid throwing exceptions in that case.

pipeline/src/ui.provider.pact.test.ts
```diff
@@ -128,25 +128,28 @@ describe('Pact Provider Verification', () => {
   });
 });
 
-// eslint-disable-next-line @typescript-eslint/require-await
 async function setupEmptyState(): Promise<void> {
   clearState();
+
+  return Promise.resolve();
```
georg-schwarz, line 134: Is it necessary?

georg-schwarz (commented):
LGTM

pipeline/src/ui.provider.pact.test.ts
felix-oq: Functions that serve as state handlers in pact need to be `async` but here the state management does not run asynchronously. I added the line such that an empty promise is returned since it is required by the return type `Promise<void>` of the function. Without it, linting would fail.

Previously for fixing the issue, linting was disabled using `// eslint-disable-next-line @typescript-eslint/require-await`.

ui/src/storage/storageRest.ts
felix-oq: I used `null` to match the type of the corresponding `item` attribute in `StorageItemView`:
https://github.com/felix-oq/ods/blob/10bc74919264fd787a3021ed0bd24832940ef207/ui/src/storage/StorageItemView.vue#L49

I'm not too familiar with Vue, so the following question arises:
Can the `null` type of the attribute also be safely replaced by `undefined`? Alternatively the `undefined` return value of `getStoredItem` could be replaced by `null` within the `mounted` function of `StorageItemView`.

pipeline/src/ui.provider.pact.test.ts
georg-schwarz: Ah alright, good to know. Thx! :)

ui/src/storage/storageRest.ts
georg-schwarz: Oh yeah makes sense. It should be `undefined` everywhere I think.

georg-schwarz (approved):

Issue #375, opened 2021-10-25
VM2 Security Alert


georg-schwarz:
https://github.com/advisories/GHSA-rjf2-j2r6-q8gr

PR #380, opened 2021-11-04
Upgraded vm2 to newest version 3.9.5

f3l1x98:
Upgrade to at least 3.9.4 as required by #375 but since newer version with Promise fix was available, this was chosen


georg-schwarz:
It might affect the exception parsing @lunedis

lunedis (approved):
Tests are passing, they are quite extensive on the error handling, so I think this is fine to merge.

PR #383, opened 2021-11-10
Fixed the bug by added key triggered update and v-if

### LINKED ISSUE #382: UI StorageView broken
The StorageViewItem of the UI does not display any data with the following error shown in the console:
image

f3l1x98:
- Added `v-if` to display generic 'No data' message if item or item.data is undefined
- Added `updateKey` variable that is either `-1` or `item.id` causing the view to be updated [https://vuejs.org/v2/api/#key](https://vuejs.org/v2/api/#key)
- Made `mounted` method async

Closes #382

### DIFF
--- a/ui/src/storage/StorageItemView.vue
+++ b/ui/src/storage/StorageItemView.vue
@@ -1,10 +1,10 @@
 <template>
   <div>
-    <div v-if="item !== null">
+    <div v-if="item != null">
       <v-card class="grey lighten-3">
-        <v-container fluid :key="updateKey">
+        <v-container fluid>
           <pre
-            v-if="item && item.data"
+            v-if="item.data !== undefined"
             style="max-height: 400px; overflow:auto; text-align: left"
           >
             {{ item.data }}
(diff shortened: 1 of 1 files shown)


ui/src/storage/StorageItemView.vue
```diff
@@ -2,10 +2,19 @@
   <div>
     <div v-if="item !== null">
```
georg-schwarz: I think this should be "item !== undefined" 

Maybe that does the trick? For me the forced update looks rather strange - it should actually work without, this is not a too special case I think.

f3l1x98: That should be done as well, but it only changed, such that the LinearProgressIndicator is displayed the whole time
image

georg-schwarz:
I looked into it because it seemed strange to me that the update is not happening...

Turns out that the bug is somewhere else: the `item` was undefined when not loaded - and Vue didn't like that and threw an error. This error led to the update mechanism being not triggered anymore.

I fixed the original bug by using `null` as default value instead of `undefined`. Very hard to track down.

georg-schwarz (approved):

PR #385, opened 2021-11-18
Added notification project to eslint workingDirs + updated package-lock

f3l1x98:
Found out, that the notification project was missing in the eslint workingDir, causing it not to find its tsconfig.json


georg-schwarz (approved):
good catch!

PR #387, opened 2021-11-22
Fix faulty mocking in provider-side CDC-Test of Pipeline

felix-oq:
This PR fixes faulty mocking that was present in the provider-side CDC-Test. The mocked behaviour of `PipelineConfigManager` for updating and deleting pipelines did not match its actual behavior in cases where there is no pipeline with the requested id.

In order to fix that problem, I modified the interactions between UI and Pipeline to ease the implementation of the mock. In particular for updating/deleting, the `PipelineConfigManager` now returns `undefined` if a pipeline with the given id does not exist and returns the updated/deleted pipeline if it exists.
Due to these interaction changes, a unit test and some integration and system tests had to be adjusted.

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -37,8 +37,8 @@ Use `npm test` to run the unit tests. There is also `nrm run watch-test` availab
 | *base_url*/configs | GET | - | PipelineConfig[] | Get all pipeline configs |
 | *base_url*/configs/:id | GET | - | PipelineConfig | Get pipeline config by id |
 | *base_url*/configs | POST | PipelineConfigDTO | PipelineConfig | Create a pipeline config |
-| *base_url*/configs/:id | PUT | PipelineConfigDTO | text | Update a pipeline config |
-| *base_url*/configs/:id | DELETE | - | text | Delete a pipeline config by id |
+| *base_url*/configs/:id | PUT | PipelineConfigDTO | PipelineConfig | Update a pipeline config |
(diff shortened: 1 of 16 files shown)


georg-schwarz:
Saw you changed response codes in some cases. What should be the case for deletion and update on unknown element? Just to have it consistent..

felix-oq:
> Saw you changed response codes in some cases. What should be the case for deletion and update on unknown element? Just to have it consistent..

For updating or deleting pipelines with an existing id, I changed the response codes from `204` to `200` because the responses now have a body that contains the updated/deleted pipeline.  `204` doesn't fit in this case anymore because the response now has content in its body.

For trying to update a pipeline with a non-existent id, I kept the `404` status code because the update operation cannot be performed when the pipeline to update is not present.

For trying to delete a pipeline with a non-existent id, I kept the response code `204` because the operation still ensures that there remains no pipeline with the given id, even though no deletion has to be performed. In that case, the body of the response remains empty because there is no pipeline to delete. The empty content of the response is resembled by the `204` status code which also makes it possible for the client to distinguish the two cases (`200` with an actual pipeline deletion and `204` where the pipeline wasn't present in the first place).

georg-schwarz:
Discussion outsourced to https://github.com/jvalue/ods/issues/388

georg-schwarz:
I think we will go with 404 in cases of a non-existing entity.

georg-schwarz (approved):

georg-schwarz (approved):

PR #390, opened 2021-11-30
image publishing add git hash to version on non-release

### LINKED ISSUE #389: CI Image Publishing
The CI on main branch should use the git hash for labelling instead of overwriting existing ones.
On releases, we should use the version numbers.

We also need to publish to the GitHub Container registry seems to be buggy with tools like ArgoCD.

georg-schwarz:
Closes #389


knusperkrone:
That is just perfect!

PR #397, opened 2022-01-03
Retention period to 1 day for all uploaded unpublished container arti…

georg-schwarz:
…facts

### DIFF
--- a/.github/workflows/ods.yml
+++ b/.github/workflows/ods.yml
@@ -45,6 +45,7 @@ jobs:
         with:
           name: adapter-artifact
           path: adapter.tar
+          retention-days: 1
 
 
   # ----------------- SCHEDULER SERVICE --------------------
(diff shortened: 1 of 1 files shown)


.github/workflows/ods.yml
```diff
@@ -620,7 +628,7 @@ jobs:
           if [[ "${{ github.ref }}" != refs/tags/* ]]
           then
             UI_VERSION=${UI_VERSION}-$(git rev-parse --short HEAD)
-          fi
+          figit
```
MBuchalik: I have never heard of `if ... figit`. Is this a typo, or does it have a special meaning?

georg-schwarz: Thx good catch! Came from my terminal dropdown that was too slow to take my keys :/

lunedis (approved):

PR #402, opened 2022-01-27
Elastic schema

knusperkrone:
@georg-schwarz 

I found this hardcoded variable for the Kubernetes instance, as I expect a "public" schema there.

Also, the anon user gets implicit created as ReadOnly User inside the entrypoint.


storage/postgrest/entrypoint.sh
```diff
@@ -1,18 +1,23 @@
...
-printf 'Waiting for %s to be ready\n' "${DATABASE_URL}"
+printf 'Waiting for %s to be ready\n' "${DATABASE_HOST}:${DATABASE_PORT}/${DATABASE_NAME}"
 
 (( max_retry=30 ))
 (( current_retry=1 ))
 
+READ_ONLY_COMMAND="GRANT CONNECT ON DATABASE ${DATABASE_NAME} TO ${PGRST_DB_ANON_ROLE};
+GRANT USAGE ON SCHEMA ${PGRST_DB_SCHEMA} TO ${PGRST_DB_ANON_ROLE};
+GRANT SELECT ON ALL TABLES IN SCHEMA ${PGRST_DB_SCHEMA} TO ${PGRST_DB_ANON_ROLE};
+GRANT SELECT ON ALL SEQUENCES IN SCHEMA ${PGRST_DB_SCHEMA} TO ${PGRST_DB_ANON_ROLE};"
```
georg-schwarz, line 17: I think this should be done in the liquibase script where the anon user is created. Or is there a specific reason to do it here?

knusperkrone: It somehow seems not not work with the liquibase script, maybe another postgresScheme error?

knusperkrone: Ah damn, it seems like, I have to change the storage-mq scheme via a env flag, not this one.

knusperkrone:
Closing as issues was on me side.

I globally set the schema to "public", which since silently failed, as "storage" is hardcoded in postgrest and liquibase service.

@georg-schwarz may considering hardcoding it into the 'mq-service' as well, to ommit such an error in the future?

